PR #50, opened 2015-01-10
Update for 1.0.0-alpha compatibility

sbward:
* Tested with `cargo test` (passing)


src/dictionary.rs
```diff
@@ -152,7 +153,7 @@ impl CFDictionary {
     pub fn get(&self, key: *const c_void) -> *const c_void {
         let value = self.find(key);
         if value.is_none() {
-            panic!("No entry found for key: {}", key);
+            panic!("No entry found for key [FIXME]"); // FIXME Not sure how to convert key to be printed
```
sbward: This is the FIXME line

Ms2ger: `{:p}` should work.

sbward:
@Ms2ger your suggestion worked, thanks.

PR #51, opened 2015-01-23
Fixed Show -> Debug

TyOverby:
`fmt::Show` was recently renamed to `fmt::Debug`

PR #52, opened 2015-01-25
Fix Show implementations for Rust nightly

bjz:


bjz:
This is backwards compatible, unlike #51, so should get us over the line. This is currently breaking downstream libs, like glutin and gfx-rs

PR #60, opened 2015-03-28
Update for stable Rust by removing the core feature

SSheldon:
It'd be great for libraries that depend on this crate if it worked on stable Rust. With the beta approaching soon, the unstable and deprecated APIs this crate uses are:
* `std::num::ToPrimitive` and `std::num::FromPrimitive`
* `std::num::Int::max_value`
* `std::str::Str::as_slice`

This pull request replaces uses of this functionality.

The changes to `CFNumber` are a breaking change, but I only found servo/rust-core-text and servo/rust-azure that would be impacted, and the fix is simple.

### DIFF
--- a/src/array.rs
+++ b/src/array.rs
@@ -178,7 +178,6 @@ extern {
 #[test]
 fn should_box_and_unbox() {
     use number::{CFNumber, number};
-    use std::num::ToPrimitive;
 
     let arr = CFArray::from_CFTypes(&[
         number(1).as_CFType(),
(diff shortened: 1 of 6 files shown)


src/dictionary.rs
```diff
@@ -101,10 +101,12 @@ impl TCFType<CFDictionaryRef> for CFDictionary {
...
-        let (keys, values): (Vec<CFTypeRef>,Vec<CFTypeRef>) =
-            pairs.iter()
-            .map(|&(ref key, ref value)| (key.as_CFTypeRef(), value.as_CFTypeRef()))
-            .unzip();
+        let mut keys = Vec::with_capacity(pairs.len());
+        let mut values = Vec::with_capacity(pairs.len());
+        for &(ref key, ref value) in pairs {
+            keys.push(key.as_CFTypeRef());
+            values.push(value.as_CFTypeRef());
+        }
```
bjz: Why did you need to remove the iterator usages?

SSheldon: `unzip` is unfortunately unstable: http://doc.rust-lang.org/std/iter/trait.Iterator.html#tymethod.unzip

Ms2ger:
https://github.com/rust-lang/rust/pull/23908 stabilized `unzip`.

SSheldon:
Convenient! I added back `unzip` @Ms2ger.

I'm waiting for another nightly that includes rust-lang/rust#23947 so I can test this, but it appears it'll work then.

SSheldon:
Tested in the latest nightly and this looks good.

r? @Ms2ger or @bjz

bjz:
r+

PR #62, opened 2015-05-06
Upgrade to rustc 1.0.0-dev (47551b574 2015-04-14)

SimonSapin:


SimonSapin:
I know nothing about Core Foundation, but this new `from_i32` method looks close enough to the existing `from_i64`.

PR #67, opened 2015-06-10
Build and test on Travis CI

frewsxcv:


frewsxcv:
Travis is currently struggling https://www.traviscistatus.com/

I'll force push when it is no longer in a state of struggling

PR #68, opened 2015-08-26
Split a -sys crate out

sfackler:
I used some reexports to make this move backwards compatible for current users.

[Review on Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/68)


metajack:
Since when did a `-sys` crate start meaning Rust bindings to a C library? When I started this naming convention it was for creates that shipped a C library. `freetype-sys` was libfreetype2 Cargo-ified, and `freetype` was the Rust bindings to Freetype.

We are not shipping CoreFoundation.framework in Cargo packaging, so this is confusing.

metajack:
cc @aturon @alexcrichton  for input on the community standard. I'll be really disappointed if we have to rename everything because the community took the wrong cue from Servo conventions.

Manishearth:
http://doc.crates.io/build-script.html#*-sys-packages

metajack:
Beautiful. :(

So now what do C packages in Cargo get called? `pkg-config` is not a solution to things that aren't easily expected to be installed or must have specific versions or for Android, etc.

I'll a little bit cranky that I spent a bunch of time working this out, asking @alexcrichton and others for advice, only to have the solution co-opted and documented as being something other than what was decided. Now I get to do all the work again if we want to maintain community norms, and apparently I get to come up with a new naming convention that someone may then reappropriate in the future for something else!

alexcrichton:
@metajack yeah the standard is for `foo-sys` crates to provide basically a header file for the C library `foo`. It's also responsible for doing anything necessary to actually link to the library `foo`, which can include things like:

* Just printing `-l` flags
* Running `pkg-config` to find the library
* Building the library from source (either vendored or perhaps downloaded)

If the library can't actually be linked for whatever reason, it's the responsibility of the build script to print an error. In that sense the convention of `foo` and `foo-sys` crates is the same, it's just focusing the responsibility of the `foo-sys` crate to link and provide definitions for the library.

Sorry if there was any miscommunication earlier, this was the intention from the get-go and something may have just been lost along the way.

sfackler:
Rebased

sfackler:
ping @pcwalton?

larsbergstrom:
I'll review this.

larsbergstrom:
Reviewed 30 of 43 files at r1, 1 of 1 files at r2, 9 of 9 files at r3.
Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/servo/core-foundation-rs/68)

sfackler:
Thanks! Could you publish to Crates.io by any chance?

metajack:
This got published.

PR #69, opened 2015-08-30
Modernization and cleanup

sfackler:
Changes are split up into a set of separate commits.

[Review on Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/69)

### DIFF
--- a/src/array.rs
+++ b/src/array.rs
@@ -9,7 +9,7 @@
 
 //! Heterogeneous immutable arrays.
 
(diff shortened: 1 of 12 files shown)


pcwalton:
Looks good, just had a couple of nits.

sfackler:
Updated, thanks for taking a look!

sfackler:
There's also some breaking updates I'd like to make (e.g. making the parameter to `TCFType` into an associated type) - not sure how you want to schedule that stuff.

PR #76, opened 2015-12-18
CFString refactor / cleanup

TimNN:
This pull request:
- removes two unnecessary casts (from `CFIndex` to `usize` and back again)
- extracts the to rust string conversion: instead of doing it on-the-fly in the `Display` implementation (and thus only supporting the indirect `ToString` way to get a rust string from a `CFString`, which _always_ allocates) this implements `From<&CFString> for Cow<str>` allowing everyone the possibility of avoiding an allocation if the `CFString` is already formatted in UTF-8.

[Review on Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/76)

### DIFF
--- a/core-foundation/src/string.rs
+++ b/core-foundation/src/string.rs
@@ -86,8 +86,8 @@ impl<'a> From<&'a CFString> for Cow<'a, str> {
                                                      false as Boolean,
                                                      buffer.as_mut_ptr(),
                                                      buffer.len().to_CFIndex(),
-                                                     &mut bytes_used) as usize;
-                assert!(chars_written.to_CFIndex() == char_len);
+                                                     &mut bytes_used);
+                assert!(chars_written == char_len);
 
                 // This is dangerous; we over-allocate and null-terminate the string (during
                 // initialization).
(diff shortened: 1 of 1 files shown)


nox:
Isn't `CFString` a UTF-16 thing?

TimNN:
As far as I know, `CFString` does not have a specific encoding. So if it's already UTF-8, we can get away without allocating, otherwise we need a buffer so the string can be encoded as UTF-8 into it.

nox:
> The CFStringGetLength function returns the total number (in terms of UTF-16 code pairs) of characters in the string.

TimNN:
Yes, but as far as I understand, this function is only used to verify that all characters in the string have been written.

All functions actually accessing the string data always specify `kCFStringEncodingUTF8`.

nox:
Oh I see. Could we use `CFStringGetCString` in the `Owned` branch?

TimNN:
We probably could, but from reading the documentation I think `CFStringGetBytes` is more appropriate: We don't care about the `NUL` terminator which `CFStringGetCString` appends and `CFStringGetBytes` can also tell us how many bytes we actually need in the output buffer (which `CFStringGetCString` cannot as far as I can tell).

nox:
Oh right.

jdm:
@TimNN Is there any reason not to merge these changes besides maintainers being unresponsive?

TimNN:
So I *thought* that there had been a rather large change to the `string` module that would obsolete / complicate the changes proposed here, but apparently that is not the case. I'm no longer working on the (hobby) project which needed these changes and was cleaning up my open PRs.

If there's still interest in getting this merged, I'd be happy to rebase the PR or let someone else take over.

jdm:
I'm in favour of merging a rebased version :)

TimNN:
Opened #220 with the rebase.

PR #82, opened 2016-06-14
Update CGNumber

LuoZijun:
update CGNumber.


core-foundation-sys/src/number.rs
```diff
@@ -50,6 +53,8 @@ extern {
     pub static kCFBooleanFalse: CFBooleanRef;
 
     pub fn CFBooleanGetTypeID() -> CFTypeID;
+    pub fn CFBooleanGetValue(boolean: CFBooleanRef ) -> bool;
```
metajack: Remove space after `CFBooleanRef` and before `)`.

core-foundation-sys/src/string.rs
```diff
@@ -188,7 +188,7 @@ pub type CFStringEncodings = CFIndex;
 //static kCFStringEncodingShiftJIS_X0213_00: CFStringEncoding = 0x0628; /* Deprecated */
 
 #[repr(C)]
-pub struct __CFString(c_void);
+pub struct __CFString(pub c_void);
```
metajack: Why does this need to be pub?

core-foundation-sys/src/string.rs
```diff
@@ -259,7 +259,12 @@ extern {
     //fn CFStringGetCharacters
     //fn CFStringGetCharactersPtr
     //fn CFStringGetCharacterFromInlineBuffer
-    //fn CFStringGetCString
+    pub fn CFStringGetCString(theString: CFStringRef, 
+                                buffer: *mut *const c_char,
```
metajack: I believe the arguments are supposed to line up in the column after the `(`.

core-foundation/src/array.rs
```diff
@@ -18,7 +18,7 @@ use std::mem;
 use base::{CFIndexConvertible, TCFType};
 
 /// A heterogeneous immutable array.
-pub struct CFArray(CFArrayRef);
+pub struct CFArray(pub CFArrayRef);
```
metajack: Why does this need to be pub?

core-foundation/src/dictionary.rs
```diff
@@ -19,7 +19,7 @@ use std::ptr;
 use base::{CFType, CFIndexConvertible, TCFType};
 
 /// An immutable dictionary of key-value pairs.
-pub struct CFDictionary(CFDictionaryRef);
+pub struct CFDictionary(pub CFDictionaryRef);
```
metajack: Why pub?

core-foundation/src/array.rs
LuoZijun: @metajack Example:

```rust
let cf_array_ref: CFArrayRef = CFArrayCreate(some_args...);
// now tuple struct need be public.
let cf_array: CFArray = CFArray(cf_array_ref);
// or just one line: 
let cf_array: CFArray = CFArray( CFArrayCreate(some_args...) );

/// And if not public ,we can't doing like this:
let cf_array_ref = cf_array.0;

```

LuoZijun: just like `pub struct CFArray(pub CFArrayRef);`

core-foundation/src/dictionary.rs
LuoZijun: just like `pub struct CFArray(pub CFArrayRef); `

core-foundation-sys/src/string.rs
LuoZijun: agree.

core-foundation-sys/src/number.rs
LuoZijun: agree

core-foundation/src/string.rs
```diff
@@ -136,10 +137,8 @@ impl CFString {
 
     /// Returns the number of characters in the string.
     #[inline]
-    pub fn char_len(&self) -> CFIndex {
-        unsafe {
-            CFStringGetLength(self.0)
-        }
+    pub unsafe fn char_len(&self) -> CFIndex {
+        CFStringGetLength(self.0)
```
KiChjang: Why is this function made unsafe as a whole? Is `CFStringGetLength` not a safe interface?

LuoZijun: @KiChjang 
This is C ffi code, so not safe.

KiChjang: Yes I understand that, but why did you remove the unsafe block inside `char_len` and instead made `char_len` unsafe?

PR #84, opened 2016-09-03
Add bindings for `CFFileDescriptor`

trlim:
This pull request adds implementation of `CFFileDescriptor` and `CFRunLoopSource` related API.

---

This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/84)

### DIFF
--- a/core-foundation/src/filedescriptor.rs
+++ b/core-foundation/src/filedescriptor.rs
@@ -81,6 +81,17 @@ impl AsRawFd for CFFileDescriptor {
     }
 }
 
+use runloop::{CFRunLoopSource};
+
+impl CFRunLoopSource {
+    pub fn from_file_descriptor(fd: &CFFileDescriptor, order: CFIndex) -> CFRunLoopSource {
+        unsafe {
+            let source_ref = CFFileDescriptorCreateRunLoopSource(kCFAllocatorDefault, fd.0, order);
+            TCFType::wrap_under_create_rule(source_ref)
+        }
+    }
+}
+
 #[cfg(test)]
 mod test {
     use super::*;
--- a/core-foundation/src/filedescriptor.rs
+++ b/core-foundation/src/filedescriptor.rs
@@ -25,7 +25,7 @@ impl CFFileDescriptor {
     pub fn new(fd: RawFd,
                closeOnInvalidate: bool,
                callout: CFFileDescriptorCallBack,
-               context: Option<&CFFileDescriptorContext>) -> CFFileDescriptor {
+               context: Option<&CFFileDescriptorContext>) -> Option<CFFileDescriptor> {
         unsafe {
             let fd_ref = CFFileDescriptorCreate(kCFAllocatorDefault,
                                                 fd,
@@ -36,7 +36,11 @@ impl CFFileDescriptor {
                                                 } else {
                                                     ptr::null()
                                                 });
-            TCFType::wrap_under_create_rule(fd_ref)
+            if fd_ref.is_null() {
+                None
+            } else {
+                Some(TCFType::wrap_under_create_rule(fd_ref))
+            }
         }
     }
 
@@ -84,10 +88,14 @@ impl AsRawFd for CFFileDescriptor {
 use runloop::{CFRunLoopSource};
 
 impl CFRunLoopSource {
-    pub fn from_file_descriptor(fd: &CFFileDescriptor, order: CFIndex) -> CFRunLoopSource {
+    pub fn from_file_descriptor(fd: &CFFileDescriptor, order: CFIndex) -> Option<CFRunLoopSource> {
         unsafe {
             let source_ref = CFFileDescriptorCreateRunLoopSource(kCFAllocatorDefault, fd.0, order);
-            TCFType::wrap_under_create_rule(source_ref)
+            if source_ref.is_null() {
+                None
+            } else {
+                Some(TCFType::wrap_under_create_rule(source_ref))
+            }
         }
     }
 }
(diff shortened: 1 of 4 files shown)


nox:
Pretty much every function in that PR that takes a raw pointer should be marked unsafe.

trlim:
I don't know if I'm getting right your point but I changed all(two) functions take raw pointer as unsafe.
But `CFRunLoopTimer::new` also seems to be marked unsafe in this sense.

trlim:
I just updated this PR with rebased/reorganized commits and some changes.
The main difference is that `CFFileDescriptor::new` now has safe signature which takes optional reference to `CFFileDescriptorContext` and doesn't require unsafe pointer anymore.

core-foundation/src/filedescriptor.rs
```diff
@@ -0,0 +1,71 @@
...
+}
+
+impl_TCFType!(CFFileDescriptor, CFFileDescriptorRef, CFFileDescriptorGetTypeID);
+
+impl CFFileDescriptor {
+    pub unsafe fn new(fd: RawFd,
+                      closeOnInvalidate: bool,
+                      callout: CFFileDescriptorCallBack,
+                      context: *const CFFileDescriptorContext) -> CFFileDescriptor {
+        let fd_ref = CFFileDescriptorCreate(kCFAllocatorDefault,
```
jdm: This can return null if there's a problem, according to [the docs](https://developer.apple.com/documentation/corefoundation/1477591-cffiledescriptorcreate?language=objc).

core-foundation/src/filedescriptor.rs
```diff
@@ -20,16 +22,30 @@ impl Drop for CFFileDescriptor {
...
+    pub fn new(fd: RawFd,
+               closeOnInvalidate: bool,
+               callout: CFFileDescriptorCallBack,
+               context: Option<&CFFileDescriptorContext>) -> CFFileDescriptor {
+        unsafe {
+            let fd_ref = CFFileDescriptorCreate(kCFAllocatorDefault,
+                                                fd,
+                                                closeOnInvalidate as Boolean,
+                                                callout,
+                                                if let Some(context) = context {
```
jdm: Let's do this instead:
```rust
let context = context.map_or(ptr::null(), |c| c as *const _);
```

core-foundation/src/filedescriptor.rs
```diff
@@ -81,6 +81,17 @@ impl AsRawFd for CFFileDescriptor {
     }
 }
 
+use runloop::{CFRunLoopSource};
+
+impl CFRunLoopSource {
+    pub fn from_file_descriptor(fd: &CFFileDescriptor, order: CFIndex) -> CFRunLoopSource {
+        unsafe {
+            let source_ref = CFFileDescriptorCreateRunLoopSource(kCFAllocatorDefault, fd.0, order);
```
jdm: This can return null according to [the docs](https://developer.apple.com/documentation/corefoundation/1477593-cffiledescriptorcreaterunloopsou?language=objc).

core-foundation/src/filedescriptor.rs
```diff
@@ -94,5 +94,98 @@ impl CFRunLoopSource {
...
+
+        // close() should succeed
+        assert_eq!(unsafe { libc::close(raw_fd) }, 0);
+    }
+
+    extern "C" fn never_callback(_f: CFFileDescriptorRef,
+                                 _callback_types: CFOptionFlags,
+                                 _info_ptr: *mut c_void) {
+        // should never be called
+        assert!(false);
```
jdm: Let's use `unreachable!()` instead.

core-foundation/src/filedescriptor.rs
```diff
@@ -94,5 +94,98 @@ impl CFRunLoopSource {
...
+    use runloop::{CFRunLoop};
+
+    #[test]
+    fn test_consumed() {
+        let path = CString::new("/dev/null").unwrap();
+        let raw_fd = unsafe { libc::open(path.as_ptr(), O_RDWR, 0) };
+        let cf_fd = CFFileDescriptor::new(raw_fd, true, never_callback, None);
+
+        assert!(cf_fd.valid());
+        cf_fd.invalidate();
```
jdm: Let's assert !cf_fd.valid().

core-foundation/src/filedescriptor.rs
```diff
@@ -94,5 +94,98 @@ impl CFRunLoopSource {
...
+    }
+
+    #[test]
+    fn test_unconsumed() {
+        let path = CString::new("/dev/null").unwrap();
+        let raw_fd = unsafe { libc::open(path.as_ptr(), O_RDWR, 0) };
+        let cf_fd = CFFileDescriptor::new(raw_fd, false, never_callback, None);
+
+        assert!(cf_fd.valid());
+        cf_fd.invalidate();
```
jdm: Let's assert !cf_fd.valid().

core-foundation/src/filedescriptor.rs
```diff
@@ -94,5 +94,98 @@ impl CFRunLoopSource {
...
+        assert_eq!(info.value, kCFFileDescriptorWriteCallBack);
+
+        info.value = 0;
+        cf_fd.disable_callbacks(kCFFileDescriptorReadCallBack|kCFFileDescriptorWriteCallBack);
+
+        cf_fd.invalidate();
+    }
+
+    extern "C" fn callback(_f: CFFileDescriptorRef, callback_types: CFOptionFlags, info_ptr: *mut c_void) {
+        assert!(info_ptr != ptr::null_mut());
```
jdm: `assert!(!info_ptr.is_null());`

core-foundation/src/filedescriptor.rs
```diff
@@ -94,5 +94,98 @@ impl CFRunLoopSource {
...
+        CFRunLoop::run_current();
+        assert_eq!(info.value, kCFFileDescriptorReadCallBack);
+
+        info.value = 0;
+        cf_fd.enable_callbacks(kCFFileDescriptorWriteCallBack);
+        CFRunLoop::run_current();
+        assert_eq!(info.value, kCFFileDescriptorWriteCallBack);
+
+        info.value = 0;
+        cf_fd.disable_callbacks(kCFFileDescriptorReadCallBack|kCFFileDescriptorWriteCallBack);
```
jdm: nit: spaces around |

trlim:
Applied changes as requested.

jdm:
@bors-servo r+
Thanks!

PR #98, opened 2017-01-31
Do not use 'links' in Cargo.toml (fixes #97)

### LINKED ISSUE #97: Stop using links
AFAIK, the `links` attribute in Cargo only exist because on some platforms, linking multiple times to a single thing is a recipe for failure. I don't think that is the case for macOS frameworks, and having Cargo fail because two core-foundation-sys versions are used in a single project is a PITA (see #96).

We should probably not use `links` if we can afford to make do without it.

Cc @tomaka for the initial paper cut, @alexcrichton for his Cargo knowledge and to confirm or infirm what I just wrote.

nox:
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/98)

### DIFF
--- a/core-foundation-sys/Cargo.toml
+++ b/core-foundation-sys/Cargo.toml
@@ -3,11 +3,10 @@ name = "core-foundation-sys"
 description = "Bindings to Core Foundation for OS X"
 homepage = "https://github.com/servo/core-foundation-rs"
 repository = "https://github.com/servo/core-foundation-rs"
-version = "0.3.0"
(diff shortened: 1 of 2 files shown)


jdm:
If someone verifies that an application relying on core-foundation-rs still builds after this change, r=me.

nox:
I checked with the cocoa examples.

@bors-servo r=jdm

jrmuizel:
I think you may need to kick the ci (travis had a bad day today).

PR #99, opened 2017-03-31
Add CFSetApplyFunction and CFSetGetCount

jcjones:
I'm needing these CFSet functions for some USB HID interactions with IOKit, so here's the external references.

They are defined in CoreFoundation:

- [CFSetApplyFunction](https://developer.apple.com/reference/corefoundation/1520450-cfsetapplyfunction?language=objc) and 
- [CFSetGetCount](https://developer.apple.com/reference/corefoundation/1520419-cfsetgetcount?language=objc)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/99)

### DIFF
--- a/core-foundation-sys/src/set.rs
+++ b/core-foundation-sys/src/set.rs
@@ -11,6 +11,8 @@ use libc::c_void;
 
 use base::{CFAllocatorRef, CFIndex, CFTypeID};
 
+pub type CFSetApplierFunction = extern "C" fn (value: *const c_void,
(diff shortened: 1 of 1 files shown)


jdm:
@metajack How did we fix the problem with core-graphics-rs last time? It looks like the same issue is happening in this repository.

metajack:
@jdm We were missing "status updates" in the webhook settings. I just added that.

PR #101, opened 2017-05-22
Additions to CFURL and CFBundle

mgoszcz2:
I propose a few additions to CFBundle and CFURL APIs. Not sure how general they are, but they would be nice to have out of the box.

The `CFURL::absolute_file_url` test is a bit big, but it demonstrates the need for `CFURL::absolute()`.

I think `CFBundle::main_bundle()` should return an `Option` since `CFBundleGetMainBundle` can return NULL but that is a breaking change and I decided not to include it.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/101)

### DIFF
--- a/core-foundation-sys/src/bundle.rs
+++ b/core-foundation-sys/src/bundle.rs
@@ -9,7 +9,8 @@
 
 use libc::c_void;
 
-use base::CFTypeID;
+use base::{CFTypeID, CFAllocatorRef};
+use url::CFURLRef;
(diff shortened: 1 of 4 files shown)


jdm:
Other than the spelling, this looks fine.

mgoszcz2:
When will you be publishing the next version?

jdm:
Could you make a PR that updates the minor version in Cargo.toml?

jdm:
Published.

PR #110, opened 2017-08-08
Optional 10.8 support

### LINKED ISSUE #109: 10.8 features on 10.7?
I'd like to add a feature to Cargo itself that requires macOS 10.8, but Cargo supports 10.7 as well.

I'm afraid that exposing `kCFURLIsExcludedFromBackupKey` (10.8+) will cause link errors on macOS 10.7.

Can it be linked using `#[linkage = "extern_weak"]`? It seems to be an unstable feature. Is it OK to add `#![feature(linkage)]` to this crate?

pornel:
Closes #109

---
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/110)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,9 @@
 # core-foundation-rs
 
 [![Build Status](https://travis-ci.org/servo/core-foundation-rs.svg?branch=master)](https://travis-ci.org/servo/core-foundation-rs)
+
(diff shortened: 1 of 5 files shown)


jdm:
@bors-servo: r+
Thanks!

PR #120, opened 2017-10-19
Support converting non-utf8 paths to CFURLs

jrmuizel:
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/120)


SimonSapin:
Does core-foundation even exist on non-OS-X?

jrmuizel:
Yeah, you can compile it for a variety of platforms. Windows is explicitly
supported. I don't think this crate supports that, but I didn't want to
preclude it from happening.

SimonSapin:
Alright. Looks good!

@bors-servo r+

PR #122, opened 2017-10-20
Implement to_path() to convert CFURLs to PathBufs.

jrmuizel:
This lets us roundtrip Paths through CFURLs

---
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/122)

### DIFF
--- a/core-foundation/src/url.rs
+++ b/core-foundation/src/url.rs
@@ -14,13 +14,18 @@ pub use core_foundation_sys::url::*;
 use base::{TCFType, CFIndex};
 use string::{CFString};
 
-use core_foundation_sys::base::{kCFAllocatorDefault, CFRelease};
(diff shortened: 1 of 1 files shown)


core-foundation/src/url.rs
```diff
@@ -77,6 +82,18 @@ impl CFURL {
         }
     }
 
+    #[cfg(unix)]
+    pub fn to_path(&self) -> PathBuf {
+        // implementing this on Windows is more complicated because of the different OsStr representation
+        unsafe {
+            let mut buf: [u8; PATH_MAX as usize] = mem::uninitialized();
+            CFURLGetFileSystemRepresentation(self.0, true as Boolean, buf.as_mut_ptr(), buf.len() as CFIndex);
```
jdm: This returns false if an error occurred, so we should make this method return None if that occurs.

PR #127, opened 2017-11-17
Add a type parameter to CFArray

jrmuizel:
This puts us on a path to greater type safety when using CFArrays

---
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/127)

### DIFF
--- a/core-foundation/src/array.rs
+++ b/core-foundation/src/array.rs
@@ -12,15 +12,40 @@
 pub use core_foundation_sys::array::*;
 pub use core_foundation_sys::base::{CFIndex, CFRelease};
 use core_foundation_sys::base::{CFTypeRef, kCFAllocatorDefault};
+use base::CFType;
 use libc::c_void;
 use std::mem;
+use std::marker::PhantomData;
 
 use base::{CFIndexConvertible, TCFType, CFRange};
 
 /// A heterogeneous immutable array.
-pub struct CFArray(CFArrayRef);
+pub struct CFArray<T = *const c_void>(CFArrayRef, PhantomData<T>);
 
-impl Drop for CFArray {
+/// A trait describing how to convert from the stored *const c_void to the desired T
+pub trait FromVoid {
+    fn from_void(x: *const c_void) -> Self;
+}
+
+impl FromVoid for u32 {
+    fn from_void(x: *const c_void) -> u32 {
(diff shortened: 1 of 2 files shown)


core-foundation/src/array.rs
```diff
@@ -12,31 +12,56 @@
...
 use base::{CFIndexConvertible, TCFType, CFRange};
 
 /// A heterogeneous immutable array.
-pub struct CFArray(CFArrayRef);
+pub struct CFArray<T = *const c_void>(CFArrayRef, PhantomData<T>);
 
-impl Drop for CFArray {
+/// A trait describing how to convert from the stored *const c_void to the desired T
+pub trait FromVoid<T> {
+    fn from_void(x: *const c_void) -> T;
```
jdm: The type parameter seems redundant when we could just use Self instead.

jrmuizel: Fixed

core-foundation/src/array.rs
```diff
@@ -12,31 +12,56 @@
...
 use base::{CFIndexConvertible, TCFType, CFRange};
 
 /// A heterogeneous immutable array.
-pub struct CFArray(CFArrayRef);
+pub struct CFArray<T = *const c_void>(CFArrayRef, PhantomData<T>);
 
-impl Drop for CFArray {
+/// A trait describing how to convert from the stored *const c_void to the desired T
+pub trait FromVoid {
+    fn from_void(x: *const c_void) -> Self;
```
nox: This should be unsafe.

core-foundation/src/array.rs
```diff
@@ -12,31 +12,56 @@
...
+
+impl FromVoid for *const c_void {
+    fn from_void(x: *const c_void) -> *const c_void {
+        x
+    }
+}
+
+impl FromVoid for CFType {
+    fn from_void(x: *const c_void) -> CFType {
+        unsafe { TCFType::wrap_under_get_rule(mem::transmute(x)) }
```
nox: This is unsound, one can pass `0xdeadbeef as *const c_void` or any other arbitrary and trigger UB.

jrmuizel:
Does https://github.com/servo/core-foundation-rs/pull/134 address this sufficiently?

PR #129, opened 2017-11-22
Higher level dictionary find/contains_key

faern:
I'm not very familiar with Core Foundation yet, so tell me if this is not a very helpful addition.

Adding versions of find and contains_key that can take any `TCFType` and automatically converts it to the raw pointer of its corresponding ref type. It looks like dictionaries will mostly contain types implementing this trait, and thus these versions of the methods can make it a bit more ergonomic to use the dictionary. The best would of course be if we could return the correct type as well, but I guess that is not possible with the available information.

Better suggestions for the names are welcome :)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/129)

### DIFF
--- a/core-foundation/src/dictionary.rs
+++ b/core-foundation/src/dictionary.rs
@@ -70,6 +70,13 @@ impl CFDictionary {
         }
     }
 
+    /// Similar to `contains_key` but acts on a higher level, automatically converting from any
+    /// `TCFType` to the raw pointer of its concrete TypeRef.
+    #[inline]
+    pub fn contains_key2<X, K: TCFType<*const X>>(&self, key: &K) -> bool {
+        self.contains_key(key.as_concrete_TypeRef() as *const c_void)
+    }
+
     #[inline]
     pub fn find(&self, key: *const c_void) -> Option<*const c_void> {
         unsafe {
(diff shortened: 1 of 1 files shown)


core-foundation/src/dictionary.rs
```diff
@@ -82,13 +89,23 @@ impl CFDictionary {
...
+    /// # Panics
+    ///
+    /// Panics if the key is not present in the dictionary. Use `find` to get an `Option` instead
+    /// of panicking.
     #[inline]
     pub fn get(&self, key: *const c_void) -> *const c_void {
-        let value = self.find(key);
-        if value.is_none() {
-            panic!("No entry found for key {:p}", key);
+        match self.find(key) {
```
KiChjang: This looks like an expanded version of `expect`.

faern: Indeed. I just kept the `panic!` existing before this PR. With expect I can't do the formatting as easily (will need an extra `format!`) and I did not want to remove the printing of the pointer. I don't see the point in printing the pointer but maybe there is one so I kept it. If not, the entire implementation can simply be
```rust
self.find(key).expect("No entry found for key")
```
To be honest I don't see the reason to have this `get` at all. Much more explicit to use `find` and have to `unwrap` yourself if you want panicking behavior.

I guess it would break some stuff, but I think it would be cleaner to get rid of this panicking version altogether and rename `find` to `get`. From my experience `get` is a more common name with dictionaries, plus it would match better with the underlying `CFDictionaryGetValueIfPresent`.

KiChjang: Ah, so there's the `format!` macro that you can use if formatting is the problem you're facing.

faern: Yes. So under the conditions that `get` should remain and that it should panic with the pointer in the message, now it probably has the simplest implementation.

PR #131, opened 2017-11-24
Add CFPropertyList and methods for casting to subclasses

faern:
https://developer.apple.com/documentation/corefoundation/cfpropertylist-rif specifies the following types as valid subclasess of a property list:
CFData, CFString, CFArray, CFDictionary, CFDate, CFBoolean, and CFNumber.

So I add a trait, `CFPropertyListSubClass`, and implement it for the valid types. Then add the `CFPropertyList` type and make it work as much as possible as the other `TCFType`s. However, it can't implement that trait, since it can't implement the static method `type_id`.

The main point of adding this was to then get to the convenience methods `to_CFPropertyList` and `downcast` that allows safe casting between the subclasses and the superclass. This is handy in calls such as `SCDynamicStoreSetValue` and `SCDynamicStoreCopyValue` where the methods either accept or return a `CFPropertyListRef`. With this addition it becomes easy to write safe wrappers around those functions and make them accept proper, constrained types.

Disclaimer: I'm not yet super familiar with CF. So I can only hope this is safe and sound, but I think it is.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/131)

### DIFF
--- a/core-foundation/src/propertylist.rs
+++ b/core-foundation/src/propertylist.rs
@@ -60,9 +60,13 @@ pub fn create_data(property_list: *const c_void, format: CFPropertyListFormat) -
 }
 
 
-/// Trait for all subclasses of CFPropertyList.
+/// Trait for all subclasses of [`CFPropertyList`].
+///
+/// [`CFPropertyList`]: struct.CFPropertyList.html
 pub trait CFPropertyListSubClass<Raw>: TCFType<*const Raw> {
-    /// Create an instance of the superclass type `CFPropertyList` for this instance.
+    /// Create an instance of the superclass type [`CFPropertyList`] for this instance.
+    ///
+    /// [`CFPropertyList`]: struct.CFPropertyList.html
     fn to_CFPropertyList(&self) -> CFPropertyList {
         unsafe { CFPropertyList::wrap_under_get_rule(self.as_concrete_TypeRef() as *const c_void) }
     }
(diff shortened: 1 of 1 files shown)


core-foundation/src/propertylist.rs
```diff
@@ -56,8 +58,142 @@ pub fn create_data(property_list: *const c_void, format: CFPropertyListFormat) -
...
+    /// # use core_foundation::propertylist::{CFPropertyList, CFPropertyListSubClass};
+    /// #
+    /// // Create a string.
+    /// let string: CFString = CFString::from_static_string("FooBar");
+    /// // Cast it up to a property list.
+    /// let propertylist: CFPropertyList = string.to_CFPropertyList();
+    /// // Cast it down again.
+    /// assert!(propertylist.downcast::<_, CFString>().unwrap().to_string() == "FooBar");
+    /// ```
+    pub fn downcast<Raw, T: CFPropertyListSubClass<Raw>>(&self) -> Result<T, ()> {
```
jdm: I lean towards returning Option instead of Result here.

core-foundation/src/propertylist.rs
```diff
@@ -56,8 +58,142 @@ pub fn create_data(property_list: *const c_void, format: CFPropertyListFormat) -
     }
 }
 
+
+/// Trait for all subclasses of CFPropertyList.
+pub trait CFPropertyListSubClass<Raw>: TCFType<*const Raw> {
```
jdm: We should be able to do `pub trait CFPropertyListSubClass: TCFType<*const Self> {` and avoid some redundancy in the implementations.

faern: Getting rid of the generic would be awesome. But I don't understand exactly how that would work. `Self` here is `CFString` or similar.

core-foundation/src/propertylist.rs
faern: Done

core-foundation/src/propertylist.rs
faern: Separate from this I'm working on trying to make the `ConcreteTypeRef` in `TCFType` a associated type instead of a generic. When/if that works out this can be greatly simplified.

jdm: `impl CFPropertyListSubClass<::data::__CFData> for ::data::CFData` duplicates the type in two places. Using the snippet I provided, the resulting impl of `impl CFPropertyListSubClass for ::data::CFData` should work just fine: https://play.rust-lang.org/?gist=b89fc9fe403069fd843f6054897f21db&version=stable

faern: Yes, but `impl U<*const u32> for u32 {}` does not hold for us.

Our `CFString` does not impl `TCFType<*const CFString>`, it implements it for `TCFType<*const __CFString>`.

faern: The existance of both `CFStringRef` and `__CFString` makes it complicated in more than one situation for me. I can't cast between pointers in the downcast method because the compiler does not know that `CFStringRef` is a raw pointer. It is, and all `*Ref` are `*const _`, but the compiler does not know. Which forces me to reference the underlying `*const __CF<Foo>` types in order to show the compiler the casts are fine. Maybe it's possible to work around it in some way, but I have not found one yet.

jdm: Oh, whoops. I didn't notice the underscores in the type name.

faern: Easy thing to do. Lots of types with very similar names here :)

PR #132, opened 2017-11-24
Reference type as associated value instead of generic

faern:
Each `CFFoo` has exactly one `CFFooRef` and which one is defined by `CFFoo`, not by its user. Thus the type of reference can be an associated type instead of a generic.

As you can see this greatly simplifies things such as calling `instance_of` and `from_CFType_pairs` as well as using `CFPropertyListSubClass`. It basically removes one generic parameter from any method dealing with `TCFType`s in some way.

This is a breaking change, so it would probably mean a `0.5.0` release if accepted. I don't know what your stance on breaking changes or version bumps is, so I don't know if you want this or not. But it would simplify the API, both here, but first and foremost in higher level crates using this.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/132)


core-foundation/src/dictionary.rs
```diff
@@ -73,8 +73,9 @@ impl CFDictionary {
     /// Similar to `contains_key` but acts on a higher level, automatically converting from any
     /// `TCFType` to the raw pointer of its concrete TypeRef.
     #[inline]
-    pub fn contains_key2<X, K: TCFType<*const X>>(&self, key: &K) -> bool {
-        self.contains_key(key.as_concrete_TypeRef() as *const c_void)
+    pub fn contains_key2<K: TCFType>(&self, key: &K) -> bool {
+        let type_ref_ptr = &key.as_concrete_TypeRef() as *const _ as *const *const c_void;
+        self.contains_key(unsafe { *type_ref_ptr })
```
faern: This implementation had to pay with a bit complexity for the reduced API complexity.

Basically, now I don't know that `K::TypeRef` is a `*const _` any longer, so I can't just cast the pointer directly. Instead I have to take the pointer *to the pointer* and then cast that and lastly dereference.

I added a unit test to make sure I got it right.

core-foundation/src/propertylist.rs
```diff
@@ -190,14 +190,18 @@ impl CFPropertyList {
...
     /// [`Box::downcast`]: https://doc.rust-lang.org/std/boxed/struct.Box.html#method.downcast
-    pub fn downcast<Raw, T: CFPropertyListSubClass<Raw>>(&self) -> Option<T> {
-        if self.instance_of::<_, T>() {
-            Some(unsafe { T::wrap_under_get_rule(self.0 as *const Raw) })
+    pub fn downcast<T: CFPropertyListSubClass>(&self) -> Option<T> {
+        if self.instance_of::<T>() {
+            let ptr = &self.0 as *const _ as *const <T as TCFType>::TypeRef;
+            unsafe {
+                let reference: &<T as TCFType>::TypeRef = &*ptr;
+                Some(T::wrap_under_get_rule(mem::transmute_copy(reference)))
```
faern: This implementation became uglier in order to get rid of the `Raw` generic. Using `mem::transmute_copy` is not very recommended maybe, but I hope this is used correctly. Here `reference` is of type `&CFFooRef`, which after the `transmute_copy` should be a `CFFooRef`. So the `transmute_copy` don't change the type, it just helps me take ownership of a dereferenced raw pointer. Which should be safe here unless I'm mistaken.

core-foundation/src/propertylist.rs
```diff
@@ -242,14 +246,14 @@ pub mod test {
     #[test]
     fn downcast_string() {
         let propertylist = CFString::from_static_string("Bar").to_CFPropertyList();
-        assert!(propertylist.downcast::<_, CFString>().unwrap().to_string() == "Bar");
-        assert!(propertylist.downcast::<_, CFBoolean>().is_none());
+        assert!(propertylist.downcast::<CFString>().unwrap().to_string() == "Bar");
+        assert!(propertylist.downcast::<CFBoolean>().is_none());
```
faern: Here one can see one of the clear benefits of this PR.

core-foundation/src/propertylist.rs
faern: Maybe a more elegant solution is possible? I would be eager to learn a better way if there is one.

jdm: Help me understand why we can't use `Some(T::wrap_under_get_rule(ptr))`? That method should accept `*const T::ConcreteTypeRef` if I'm not mistaken.

faern: This is of course depending on the assumption that all `TCFType::TypeRef`s have the same size as `*const c_void`. This is true for all the possible subclasses of `CFPropertyList` at the moment and probably forever, but nothing is enforcing it.

faern: If I add a second associated type, `TCFType::RawTypeRef` (or similar name), that we set to the corresponding `__CFFoo` type. Then I can refer to `*const Self::RawTypeRef` in places and do proper casts. From some early tinkering around it looks like that can get rid of all these strange solutions.

faern:
Now I'm much more satisfied with the solution to this. By adding a trait to the sys crate that provide helper functions for const raw pointers (all CFref-types) I'm now able to know more about `TCFType::ConcreteTypeRef` and thus perform conversion without any `transmute_copy` or other ugly pointer tricks.

faern:
I have investigated a little bit how much this change would break published crates depending on the 0.4 version of `core-foundation`. I tried the following with the following results:

No change required:
* cocoa_image
* core-graphics
* glutin
* offscreen_gl_context

Some change required, but very little:
* core-text
* io-surface
* font-loader

core-foundation-sys/src/error.rs
```diff
@@ -15,7 +15,7 @@ use string::CFStringRef;
 #[repr(C)]
 pub struct __CFError(c_void);
 
-pub type CFErrorRef = *mut __CFError;
+pub type CFErrorRef = *const __CFError;
```
faern: I'm not sure if this was `*mut` for a reason I have missed. But it was the only CF-ref typedef which was not `*const` so my guess is it was a typo. All tests of this crate and all dependant crates I tried work fine after this change.

faern:
@jdm Indeed. I like how much it simplifies. But @nox did not like that it breaks stuff, which it does. However updating the dependencies to work again is quite easy.

nox:
> However updating the dependencies to work again is quite easy.

It is easy but very fastidious, so I'll not let this PR land without all the others being ready too, including a PR to update them in Servo.

faern:
@nox If it's released as `0.5.0` merging and publishing would not break anything because of semver. But I fully understand that you don't want to end up in a messy situation where only a few things is updated.

nox:
> @nox If it's released as 0.5.0 merging and publishing would not break anything because of semver. But I fully understand that you don't want to end up in a messy situation where only a few things are updated.

The issue arises when we start needing things that are unrelated to your changes, and 0.5.0 didn't land yet in servo/servo, so we must continue to maintain a 0.4 branch to get the unrelated changes in tree.

Of course, we should give you a hand and not tell you to do all the bumps.

faern:
@nox I'll be happy to help upgrade the dependencies! But I'm not sure which they are. All the ones I can see are the ones listed under https://crates.io/crates/core-foundation/reverse_dependencies But I suspect that is not all of it.

If you list them here I can start submitting PRs to them. Given that you want this feature that is.

nox:
Looking at the 0.2 to 0.3 bump [commit](https://github.com/servo/servo/commit/db3e5be878a8d41eb7c4e1be1d15a294cf5c3599) gives an idea what to do.

faern:
@nox @jdm What are your thoughts on the name `ConcreteTypeRef`? I think it's long and a bit redundant. "Concrete" is already implicit by the fact that each type must state the concrete type when they implement the trait. Having "Type" in a type is kind of a special case here where we mimic names from another API that we can't control, so not that bad, but works without IMO.

How about just calling it `Ref`? `TCFType::Ref`, or in another example context: `<T as TCFType>::Ref`. Would be more ergonomic to use and IMO conveys the same information.

If not, how about getting rid of "Concrete" and calling it `TypeRef`?

jrmuizel:
@nox, @jdm ping

faern:
@jrmuizel I didn't have time to help update servo and all dependencies yet. I still have some hope I will get to that though.

core-foundation/src/base.rs
```diff
@@ -146,12 +150,14 @@ pub trait TCFType<ConcreteTypeRef> {
 
     /// Returns true if this value is an instance of another type.
     #[inline]
-    fn instance_of<OtherConcreteTypeRef,OtherCFType:TCFType<OtherConcreteTypeRef>>(&self) -> bool {
-        self.type_of() == <OtherCFType as TCFType<_>>::type_id()
+    fn instance_of<OtherCFType: TCFType>(&self) -> bool {
+        self.type_of() == <OtherCFType as TCFType>::type_id()
```
jdm: We should be able to use `OtherCFType::type_id()` now.

core-foundation/src/propertylist.rs
```diff
@@ -151,10 +151,8 @@ impl CFPropertyList {
 
     /// Returns true if this value is an instance of another type.
     #[inline]
-    pub fn instance_of<OtherConcreteTypeRef, OtherCFType: TCFType<OtherConcreteTypeRef>>(
-        &self,
-    ) -> bool {
-        self.type_of() == <OtherCFType as TCFType<_>>::type_id()
+    pub fn instance_of<OtherCFType: TCFType>(&self) -> bool {
+        self.type_of() == <OtherCFType as TCFType>::type_id()
```
jdm: We should be able to do `OtherCFType::type_id()` now.

core-foundation/src/propertylist.rs
```diff
@@ -190,14 +188,17 @@ impl CFPropertyList {
...
     ///
     /// [`CFPropertyList`]: struct.CFPropertyList.html
     /// [`Box::downcast`]: https://doc.rust-lang.org/std/boxed/struct.Box.html#method.downcast
-    pub fn downcast<Raw, T: CFPropertyListSubClass<Raw>>(&self) -> Option<T> {
-        if self.instance_of::<_, T>() {
-            Some(unsafe { T::wrap_under_get_rule(self.0 as *const Raw) })
+    pub fn downcast<T: CFPropertyListSubClass>(&self) -> Option<T> {
+        if self.instance_of::<T>() {
+            unsafe {
+                let subclass_ref = <T as TCFType>::Ref::from_void_ptr(self.0);
```
jdm: We should be able to do `T::Ref::from_void_ptr(self.0)`.

core-foundation-sys/src/base.rs
```diff
@@ -80,6 +80,23 @@ pub struct CFAllocatorContext {
     pub preferredSize: CFAllocatorPreferredSizeCallBack
 }
 
+/// Trait for all types which are Core Foundation reference types.
```
jdm: I think this commit should be reordered to come before 3688803.

faern:
@jdm My macOS dev environment was borked, so I could not build Servo. Thus I never came around to upgrade the dependencies from Servo down to here. Would be awesome to get it in though, not sure how to fix that dev env really...

core-foundation/src/propertylist.rs
faern: Done

core-foundation/src/base.rs
faern: Done

core-foundation-sys/src/base.rs
faern: Hmm. With "this commit" I assume you mean 0fc6b62f266203f3466e79d5c31a0b7cb91980f8 ? If so, it does come before 3688803

faern:
The travis build fails with a serious crash. It might be some bug I introduced, but it might also be something with Travis, since I can't reproduce it locally. Before I dive into this problem, does anyone know if this has show up before and/or is a known problem with this crate, or if it's something I likely introduced? @jdm ?

jdm:
I don't see any issue filed for it, but it also seems unlikely that the changes in this PR would have caused it.

faern:
Commenting out the `wait_200_milliseconds` test makes it not crash. But I fail to see how this PR would create a problem with that test. And I can't reproduce it locally.

faern:
I added some good old println debugging. To see if I can pinpoint approximately what crashes. But since Travis takes over an hour for each macOS build job to start this will take a while... Very smooth change->test iteration loop :P

faern:
Now it died with
`thread 'runloop::test::wait_200_milliseconds' panicked at 'assertion failed: now - previous_now > 0.19 && now - previous_now < 0.21', src/runloop.rs:216:9`.

So I'm suspecting the error was simply that the allowed time window was too narrow and that travis got slower or something. Would explain why the error manifested itself on a machine crowded with build VMs, but not a bare metal laptop.

jrmuizel:
Note: we should try to include https://github.com/servo/core-foundation-rs/pull/135 in the 0.5 release as I believe it's also a breaking change.

faern:
@jrmuizel That might be nice to get in as well then. Yes, it's breaking since it removes `CFDictionary::set_value`. This PR and that PR will break each other, so whichever one is merged last must be changed in order to work.

faern:
The timing issue that triggered the test error has now been "fixed". Fixed in the sense that it should show up as a failed test if it fails again. And fixed as in the allowed time window has been expanded from 19-21 ms to 19-30 ms.

As you can now see in the travis runs:
`wait_200_milliseconds, elapsed: 0.2741580009460449`
It does actually take ~27 ms now, so there is no way it could fit in the previous timing window. Even the new limit of 30 is likely going to error out a few times now and then.

Since this fix is unrelated to this PR and this PR is growing large, I split it out to a separate PR. Would be nice to have that one merged first: #141

faern:
Ping @nox and @jdm. I think this PR is ready for merge and a new release. All other PRs following this all the way up to Servo are ready, see https://github.com/servo/servo/pull/19759

Maybe look at #135 for another breaking change first? If you are interested in getting that in before the release.
Also, merging #141 first will fix the failing test.
Lastly, #142 would be cool, but should not be breaking, so can be done later. If you want it at all that is.

jdm:
As soon as the other PRs finish merging, we're ready to start rolling this boulder downhill.

faern:
@jdm Cool. Yeah, the merging of the other PRs has not been super smooth the last few days. Hopefully Travis stays a bit more stable now :P
Also, I'll only be online like 3-4 more hours today. So depending on how long time it will take to get started, but there will be a fair amount of updating (removing temp commits etc.) in each subsequent PR.

faern:
@jdm At the moment it's just wasting time, trying to merge the workspace PR. It will fail a test since the test fix PR is not merged :/

faern:
@jdm I rebased and fixed the merge conflicts from the other PRs.

faern:
@jdm I have noticed that glutin uses the now removed `CFDictionary::set_value`. I'm looking into fixing that. Unrelated to this PR. But related to releasing this crate and upgrading Servo. We might need to fix that with a change in this crate.

core-foundation/src/array.rs
```diff
@@ -92,11 +92,6 @@ impl<T> CFArray<T> {
         }
     }
 
-    #[deprecated(note = "please use `as_untyped` instead")]
-    pub fn to_untyped(self) -> CFArray {
```
faern: I realize now that my previous move to deprecate `to_untyped` and add `as_untyped` was not exactly making things more correct than before, if one should follow the API guidelines to the letter.
This removed `to_untyped` is a consuming convert (`owned -> owned (non-Copy types)`), so it should have been named `into_untyped` with that signature. So the old name was slightly "wrong".
On the other hand, the new `as_untyped` only borrows `&self`, but gives out an owned object (`borrowed -> owned (non-Copy types)`), so it should be named `to_untyped`, and not `as_` since that is for conversions `borrowed -> borrowed`.

Do you think I should change the name of the method that will be present in the `0.5.0` release? Or is that too much nitpicking and too much changing the API back and forth? If it should be changed this would be the correct signature:
```rust
pub fn to_untyped(&self) -> CFArray { ... }
```
In line with all the new `downcast_into` methods, we could also add a consuming one that avoids bumping the retain count:
```rust
pub fn into_untyped(self) -> CFArray { ... }
```

jdm: Those seem like worthwhile changes.

faern: Check. I'll add a commit for that. I'll push it after my other PR is merged, as this one will need rebasing/manual conflict resolution after that one anyway.

faern: Done.

faern:
Finally! \o/

PR #139, opened 2017-11-28
Fix bug in CFArray::to_untyped that made the retain count wrong

faern:
The following code segfaults on the `0.4.5` release:
```rust
CFArray::<CFBoolean>::from_CFTypes(&[]).to_untyped().to_untyped().to_untyped()
```
Because `to_untyped` don't increment the retain count, but when the typed array it's created from goes out of scope that one will decrement it.

And since we need to wrap under the get rule (call `CFRetain`) anyway I did not see any reason to have it as a consuming `to_` method. It's more flexible as a borrowing `as_` method.

Since you probably don't want to cause API breakage I kept the old version but fixed the bug and added a deprecation notice on it.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/139)


jdm:
Let's increase the patch version number and publish a new version as part of this PR. Can we add a test, too?

faern:
Sure. I'll fix that! Just a moment.

faern:
@jdm Done! I also fixed another bug with `CFType::instance_of`. At least I think it was a bug. @nox thought so too from a short description on IRC.

faern:
I realize I only tested `as_untyped`. As long as we keep the `to_untyped` version we might want to test that as well maybe? Adding to same test...

PR #141, opened 2018-01-14
Fix timing sensitive test

faern:
This problem was found when I tried to debug #132 

But since that PR grew so large, and this fix was very unrelated to the work in that PR I split it out to a separate one.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/141)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ os: osx
 script:
   - cd core-foundation
   - cargo build --verbose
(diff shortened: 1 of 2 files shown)


faern:
@jdm  :( Broken travis.

faern:
@jdm Seems Travis is a bit more healthy now. Try again? :)

PR #135, opened 2017-11-27
Add CFMutableDictionary type.

djg:
Added CFMutableDictionary that presents the mutable dictionary
functions of CF. Tries to present a similar interface as CFDictionary.

Fixes issue #93.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/135)

PR #142, opened 2018-01-14
Add consuming cheap cast

faern:
There are a lot of places where we increment and decrement the retain count when we would not have to. For example, a very common use-case with PropertyLists is to just directly cast them to their correct sub type and then throw the PropertyList away (because there is nothing you can do with the PropertyList instance directly.)

The reason we do this, I guess, is mostly because all `Drop` implementations force a retain count decrement. So we have to keep incrementing it to keep it correct. But if we add a flag to types that can be used to disable the decrement, then we can also get rid of the increment when we virtually just want to cast a pointer between two types.

This PR implements this by adding a boolean flag to all CF structs that is used by the `Drop` impl to determine if it should `CFRelease` the instance or not. Then this flag is modified in appropriate places where we cast to other types and consume ourselves.

I added some benchmarks locally. I did not want to commit them since benchmarking only works on nighty, and I did not feel like adding a feature flag for it etc. Anyhow, the following benchmarks:
```rust
#[bench]
fn bench_before(b: &mut Bencher) {
   let string = CFString::from_static_string("Bar");
    b.iter(|| unsafe {
        string.clone()
            .to_CFPropertyList()
            .downcast::<_, CFString>()
            .unwrap()
            .disable_release();
    })
}

#[bench]
fn bench_after(b: &mut Bencher) {
    let string = CFString::from_static_string("Bar");
    b.iter(|| unsafe {
        string.clone()
            .into_CFPropertyList()
            .downcast_into::<_, CFString>()
            .unwrap()
            .disable_release();
    })
}
```
Given the following results:
```
test propertylist::test::bench_after         ... bench:          16 ns/iter (+/- 3)
test propertylist::test::bench_before        ... bench:          73 ns/iter (+/- 14)
```
The final call to `disable_release()` in the benchmarks are to prevent the type coming out of the `unwrap()` from calling `CFRelease` on drop, and thus avoid measuring that. I want to try to measure only `{to,into}_CFPropertyList`, `downcast` and `downcast_into`. The `clone` is needed because the `into` version requires it and it would then not be fair to not have it on both.

I would say that speeding it up by a factor of ~4 can be well worth it? (Well, that 4 is of course no good measurement of anything, microbenchmarks etc. etc.. The benchmark mostly shows that a lot can be gained from avoiding some retain counting.)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/142)

### DIFF
--- a/core-foundation/src/array.rs
+++ b/core-foundation/src/array.rs
@@ -47,9 +47,7 @@ unsafe impl FromVoid for CFType {
 
 impl<T> Drop for CFArray<T> {
     fn drop(&mut self) {
-        unsafe {
-            CFRelease(self.as_CFTypeRef())
-        }
+        unsafe { CFRelease(self.as_CFTypeRef()) }
     }
 }
 
--- a/core-foundation/src/array.rs
+++ b/core-foundation/src/array.rs
@@ -207,27 +207,25 @@ mod tests {
                                         n4.as_CFTypeRef(),
                                         n5.as_CFTypeRef()]);
 
-        unsafe {
-            let mut sum = 0;
+        let mut sum = 0;
 
-            let mut iter = arr.iter();
-            assert_eq!(iter.len(), 6);
-            assert!(iter.next().is_some());
-            assert_eq!(iter.len(), 5);
+        let mut iter = arr.iter();
+        assert_eq!(iter.len(), 6);
+        assert!(iter.next().is_some());
+        assert_eq!(iter.len(), 5);
 
-            for elem in iter {
-                let number: CFNumber = TCFType::wrap_under_get_rule(mem::transmute(elem));
-                sum += number.to_i64().unwrap()
-            }
-
-            assert!(sum == 15);
-
-            for elem in arr.iter() {
-                let number: CFNumber = TCFType::wrap_under_get_rule(mem::transmute(elem));
-                sum += number.to_i64().unwrap()
-            }
-
-            assert!(sum == 30);
+        for elem in iter {
+            let number: CFNumber = elem.downcast::<_, CFNumber>().unwrap();
+            sum += number.to_i64().unwrap()
         }
+
+        assert!(sum == 15);
+
+        for elem in arr.iter() {
+            let number: CFNumber = elem.downcast::<_, CFNumber>().unwrap();
+            sum += number.to_i64().unwrap()
+        }
+
+        assert!(sum == 30);
     }
 }
--- a/core-foundation/src/base.rs
+++ b/core-foundation/src/base.rs
@@ -37,6 +37,30 @@ declare_TCFType!{
 }
 
 impl CFType {
+    /// Try to downcast the `CFType` to a subclass. Checking if the instance is the
+    /// correct subclass happens at runtime and `None` is returned if it is not the correct type.
+    /// Works similar to [`Box::downcast`] and [`CFPropertyList::downcast`].
+    ///
+    /// # Examples
+    ///
+    /// ```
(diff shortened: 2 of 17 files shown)


faern:
Other than the cheap `into` casts this PR also adds the normal `downcast` to `CFType` making it easy to downcast it directly to a subtype. As you can see this makes scenarios such as is the `CFArray` tests much easier to write (and without need for `unsafe`)

jrmuizel:
I haven't looked at this in detail so don't understand it yet but I'm not really a fan of growing the size of the CFTypes

jrmuizel:
Can't we just transmute between Ref types to avoid Drop?

jrmuizel:
Or rather transmute the wrapper types?

jrmuizel:
Or perhaps add a method on the wrapper type that returns the Ref type but doesn't drop()? Most wrapper types could implement this using transmute to avoid the drop.

faern:
@jrmuizel I guess it would be possible to use `mem::forget(self)` in the consuming methods that are now utilizing `disable_release` instead. I just felt this was a bit more general and safe though, since a `mem::forget(self)` would rely on the fact that the drop *only* did the release. If any type introduced any other drop logic it would also not run that. But I guess we might be able to assume avoiding drop only does the release after all, since I now moved the drop impl to the macro, so all CF types automatically have the same code there. Except the special ones who does not yet use `declare_TCFType`.

What are your arguments against growing the structs by one bool?

faern:
I have now pushed a commit that remove `disable_release` and instead uses `mem::forget(self)` to avoid the drop impl and thus the `CFRelease` call. This leaves all structs at their original size (one pointer). If people agree this is a preferred way I can squash this last fix into the main commit doing this change.

jrmuizel:
> What are your arguments against growing the structs by one bool?
None of my arguments are amazing, but here they are:
- it causes us to lose what's mostly a zero-cost abstraction
- it has a small increase of the overhead of drop()
- it means that arrays of CFTypes use up twice as much space
- it means that we don't have any chance of transmuting the void* array that CFArrayGetValues produces into the wrapper types.

faern:
Yes, keeping them pointer sized is probably very desirable. Especially since it was so easy to achieve the same results without adding the bool.

core-foundation/src/array.rs
```diff
@@ -218,14 +216,14 @@ mod tests {
...
-                let number: CFNumber = TCFType::wrap_under_get_rule(mem::transmute(elem));
+                let number: CFNumber = TCFType::wrap_under_get_rule(mem::transmute(elem.as_concrete_TypeRef()));
                 sum += number.to_i64().unwrap()
             }
 
             assert!(sum == 15);
 
             for elem in arr.iter() {
-                let number: CFNumber = TCFType::wrap_under_get_rule(mem::transmute(elem));
+                let number: CFNumber = TCFType::wrap_under_get_rule(mem::transmute(elem.as_concrete_TypeRef()));
```
jdm: Why are these changes necessary?

faern: Ehm.. They are not, any longer. they were before when I actually changed the structs to contain a bool, then the transmute did not work because the types were of different size. I'll remove it! It's removed in the commit after anyway, but anyhow.

faern:
Optimally this should be rebased when #135 is merged and updated, since that PR add's a type that does not make use of the macro here.

faern:
PR #84 also adds a type without making use of the `declare_TCFType!` macro. So if that one is merged and I'm not away until this gets auto merged as well then I'll rebase this and fix that.

PR #143, opened 2018-01-14
[RFC] Make the implementations of FromVoid more generic which gives better ergonomics

jrmuizel:
This is probably a breaking change so would be good to include with 0.5 changes.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/143)


nox:
Can you reorder the commits, or at least not put the bumping commit in the middle of the others? Thanks.

jrmuizel:
All but the last of the commits are from https://github.com/servo/core-foundation-rs/pull/132

jrmuizel:
I needed to drop the re-implementation of get_values() because we don't have a way to implement it only for types that that can be transmuted from const void *

nox:
@jrmuizel Oh never mind then, didn't realise this was blocked on another PR.

core-foundation/src/array.rs
```diff
@@ -33,15 +32,15 @@ unsafe impl FromVoid for u32 {
...
+        mem::transmute(x)
     }
 }
 
-unsafe impl FromVoid for CFType {
-    unsafe fn from_void(x: *const c_void) -> CFType {
-        TCFType::wrap_under_get_rule(mem::transmute(x))
+unsafe impl<T: TCFType> FromVoid for T where T::Ref: FromVoid {
+    unsafe fn from_void(x: *const c_void) -> Self {
+        TCFType::wrap_under_get_rule(T::Ref::from_void(x))
```
faern: Is it maybe a bit risky to assume any void pointer should be wrapped under the get rule? I guess that depend on how one obtain the pointer? I might be mistaken how it's used, but that is my impression.

A possibly safer way would be to force the user to do what I do in downcast: https://github.com/servo/core-foundation-rs/pull/143/files#diff-e49ae9facf583e1f0d73ba355079e544R199
Where they must convert their void pointer to an appropriate ref type and then apply one of the ordinary constructors `wrap_under_get_rule` or `wrap_under_create_rule`.

faern:
@nox That other PR is, however, not blocked. It's ready for merge. And I just rebased it to make the version bump the last commit :)

jrmuizel:
CFArrays can be used for types other than CFTypes. For example, CGFontCopyTableTags returns a CFArray of u32.

core-foundation/src/array.rs
jrmuizel: I'm not sure I understand how that would work. from_void is only used internally in array.rs and is unsafe. My understanding is that CFArray contents would always be accessed with wrap_under_get_rule.

nox: > Is it maybe a bit risky to assume any void pointer should be wrapped under the get rule? I guess that depend on how one obtain the pointer? I might be mistaken how it's used, but that is my impression.

@faern It's never "more" unsafe to wrap anything under the get rule. Worst case, it will retain a value too much and leak it. I wouldn't worry about this too much for now, given we have no notion of returning a borrow of an already-retained value.

faern: True, it's not unsafe to leak. Bad wording on my part I guess. But maybe the documentation should at least state that the retain count will be incremented on this operation? :)

jrmuizel: I wonder if we should actually change CFArray to return an iterator of references instead of values...

PR #145, opened 2018-01-21
Various fixes and rearrangements

faern:
A couple of things that I found easy to do that would improve the readability and usability of the crate in various ways. Some might be more subjectively better rather than objectively better. I clogged some unrelated changes into the same PR since the merge process takes ages with Travis being this slow. Anyway, they all fit under the "cleanup" category :) Tell me if you want them split into multiple PRs.

* Export the sys part of the error module. All other modules do this. I guess it was missed?

* Fix so the macros are usable from other crates without having to depend on `core-foundation-sys` or import `CFRelease` into the scope where they are used.

* The tests in `lib.rs` were only testing the dictionary implementations, so I moved them to `dictionary.rs`

* I found it a bit undiscoverable that `CFRunLoopSource` define a method inside of `filedescriptor.rs`. At the same time it's not ideal to call the ffi-methods of the filedescriptor from runloop.rs. This rearrangement solves both of those problems by exposing `CFFileDescriptor::to_run_loop_source` that can be used from `runloop.rs`.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/145)


jdm:
@bors-servo r+
Thanks for doing this cleanup! These are nice improvements.

PR #146, opened 2018-01-21
Make info_dictionary return mutable dictionary

faern:
The problem is that glutin sets a value in an **immutable** dict here: https://github.com/servo/glutin/blob/servo/src/api/cocoa/mod.rs#L403. That does not work any longer after #135, since `set_value` is gone.

`CFBundleGetInfoDictionary` returns a `CFDictionaryRef` (not a `CFMutableDictionaryRef`) and the docs on that function does not mention the returned dictionary being mutable. So if going by that information, this PR changes into something wrong. But if the code in glutin is correct (which I know nothing about, only have to assume), then it can obviously set a value in the dict. So is it mutable?

One way to solve it is in the way done in this PR at the moment. Because then glutin can continue to set the value in the dict. But if we think the returned dict should be immutable in general (as the Apple docs seems to indicate?), then we could keep this crate as is, but instead make glutin step down one level and call `CFDictionarySetValue` unsafely by itself.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/146)

### DIFF
--- a/core-foundation/src/bundle.rs
+++ b/core-foundation/src/bundle.rs
@@ -14,7 +14,7 @@ use core_foundation_sys::base::{CFRelease, kCFAllocatorDefault};
 
 use base::TCFType;
 use url::CFURL;
-use dictionary::CFDictionary;
+use dictionary::CFMutableDictionary;
 
 
 declare_TCFType!{
@@ -42,7 +42,7 @@ impl CFBundle {
         }
     }
 
-    pub fn info_dictionary(&self) -> CFDictionary {
+    pub fn info_dictionary(&self) -> CFMutableDictionary {
         unsafe {
             let info_dictionary = CFBundleGetInfoDictionary(self.0);
             TCFType::wrap_under_get_rule(info_dictionary)


faern:
Another way to solve it would be to continue let `CFBundle::info_dictionary` return an immutable `CFDictionary`, as it likely should(?). But then add an **unsafe** `CFDictionary::to_mutable(&self) -> CFMutableDictionary`. That could be used to convert any dictionary to a mutable if you know what you are doing.

But I guess someone with better understanding of core foundation would have to fill in here.

jrmuizel:
In this case I think we can just drop this code from servo/glutin. It looks like it was added before https://github.com/servo/servo/pull/11899 and was a hack to set the app name without having a .app

It's not unlikely that we'll run into this kind of thing elsewhere and I feel like CFDictionary::to_mutable(&self) will probably be the best solution but we can probably wait until we encounter it again.

jdm:
I agree with @jrmuizel's conclusions.

PR #147, opened 2018-01-23
Add the ability to borrow values from the array. 

jrmuizel:
This is a replacement for #143. It allows iteration while avoiding the reference count churn of the previous approach. It uses a TCFTypeBorrow (could probably use a better name) to hold a sort of reference to the array data. I'm interested in people's thoughts and opinions on this approach.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/147)

### DIFF
--- a/core-foundation/src/array.rs
+++ b/core-foundation/src/array.rs
@@ -12,36 +12,52 @@
 pub use core_foundation_sys::array::*;
 pub use core_foundation_sys::base::{CFIndex, CFRelease};
 use core_foundation_sys::base::{CFTypeRef, kCFAllocatorDefault};
-use base::CFType;
 use libc::c_void;
 use std::mem;
+use std::mem::ManuallyDrop;
 use std::marker::PhantomData;
+use std;
+use std::ops::Deref;
+use std::fmt::{Debug, Formatter};
 
-use base::{CFIndexConvertible, TCFType, CFRange};
+use base::{CFIndexConvertible, TCFType, TCFTypeRef, CFRange};
 
 /// A heterogeneous immutable array.
 pub struct CFArray<T = *const c_void>(CFArrayRef, PhantomData<T>);
 
+/// A reference to an element inside the array
+pub struct ItemRef<'a, T: 'a>(ManuallyDrop<T>, PhantomData<&'a T>);
+
+impl<'a, T> Deref for ItemRef<'a, T> {
+    type Target = T;
+
+    fn deref(&self) -> &T {
+        &self.0
+    }
+}
+
+impl<'a, T: Debug> Debug for ItemRef<'a, T> {
+    fn fmt(&self, f: &mut Formatter) -> Result<(), std::fmt::Error> {
+        self.0.fmt(f)
+    }
+}
+
 /// A trait describing how to convert from the stored *const c_void to the desired T
 pub unsafe trait FromVoid {
-    unsafe fn from_void(x: *const c_void) -> Self;
+    unsafe fn from_void<'a>(x: *const c_void) -> ItemRef<'a, Self> where Self: std::marker::Sized;
 }
 
 unsafe impl FromVoid for u32 {
-    unsafe fn from_void(x: *const c_void) -> u32 {
-        x as usize as u32
+    unsafe fn from_void<'a>(x: *const c_void) -> ItemRef<'a, Self> {
+        // Functions like CGFontCopyTableTags treat the void*'s as u32's
+        // so we convert by casting directly
+        ItemRef(ManuallyDrop::new(x as u32), PhantomData)
     }
 }
 
-unsafe impl FromVoid for *const c_void {
-    unsafe fn from_void(x: *const c_void) -> *const c_void {
-        x
-    }
-}
-
-unsafe impl FromVoid for CFType {
-    unsafe fn from_void(x: *const c_void) -> CFType {
-        TCFType::wrap_under_get_rule(mem::transmute(x))
+unsafe impl<T: TCFType> FromVoid for T {
+    unsafe fn from_void<'a>(x: *const c_void) -> ItemRef<'a, Self> {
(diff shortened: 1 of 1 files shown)


jdm:
I like this approach. It doesn't appear to have any downsides that I can see.

core-foundation/src/array.rs
```diff
@@ -33,15 +34,47 @@ unsafe impl FromVoid for u32 {
...
+        x as *const T
     }
 }
 
-unsafe impl FromVoid for CFType {
-    unsafe fn from_void(x: *const c_void) -> CFType {
-        TCFType::wrap_under_get_rule(mem::transmute(x))
+unsafe impl<T: TCFType> FromVoid for T where T::Ref: FromVoid {
+    unsafe fn from_void(x: *const c_void) -> Self {
+        TCFType::wrap_under_get_rule(T::Ref::from_void(x))
```
faern: `TCFType::Ref` is already forced to implement `TCFTypeRef`, which has a `from_void_ptr` method. So you could remove `where T::Ref: FromVoid` and use `T::Ref::from_void_ptr(x)` instead. Would work the same, but would be slightly simpler/more minimal maybe?

core-foundation/src/array.rs
```diff
@@ -33,15 +34,47 @@ unsafe impl FromVoid for u32 {
...
+    }
+}
+/// A trait describing how to convert from the stored *const c_void to the desired T
+pub unsafe trait FromVoidBorrow {
+    unsafe fn from_void_borrow<'a>(x: *const c_void) -> TCFTypeBorrow<'a, Self> where Self: std::marker::Sized;
+}
+
+unsafe impl FromVoidBorrow for u32 {
+    unsafe fn from_void_borrow<'a>(x: *const c_void) -> TCFTypeBorrow<'a, Self> {
+        TCFTypeBorrow(x as usize as u32, PhantomData)
```
faern: I do not understand this casting. Maybe I'm just too tired. There is no dereferencing of the pointer, so would the resulting `u32` not contain the pointer address? Quite likely a truncated version of it as well.

core-foundation/src/array.rs
jrmuizel: I went with this approach

core-foundation/src/array.rs
jrmuizel: Yes, it's intentionally not dereferencing it. I've added a comment to explain it more and dropped the 'as usize' as it doesn't seem to be necessary.

jrmuizel:
I've cleaned this up by renaming TCFTypeBorrow to ItemRef, replacing FromVoid with FromVoidBorrow and making get() return an ItemRef instead of having a separate get_borrow()

core-foundation/src/array.rs
```diff
@@ -12,36 +12,51 @@
...
+        ItemRef(x as u32, PhantomData)
     }
 }
 
-unsafe impl FromVoid for CFType {
-    unsafe fn from_void(x: *const c_void) -> CFType {
-        TCFType::wrap_under_get_rule(mem::transmute(x))
+unsafe impl<T: TCFType> FromVoid for T {
+    unsafe fn from_void<'a>(x: *const c_void) -> ItemRef<'a, Self> {
+        ItemRef(TCFType::wrap_under_create_rule(T::Ref::from_void_ptr(x)), PhantomData)
```
faern: Won't the drop of the constructed `T` do a `CFRelease` and make the retain count wrong? If you construct a new `T` from a void pointer, should you not do it under the get rule?

Could you add a test that checks retain counts are correct before and after borrowing values from the array?

Also, nitpicking: `T::wrap_under_...` should work as well? Making it a bit more clear which type we create here.

jrmuizel: Yep. I've messed up the reference counts here. I need to think about how to fix it. wraping with the get rule will increase the retain count which defeats the point of this patch.

faern: The underlying ffi function, `CFArrayGetValueAtIndex` specifies that values returned should follow the get rule. So if you would use `CFArray` directly in Objective-C you would likely bump the retain count? So is there any reason we should be allowed not to? If we convert the value into a CF object and hold a reference to that then should the retain count not reflect how many handles there are to that object?

faern:
@jrmuizel Would it not be nice to get rid of the possible panic in `CFArray::get`? Replace it with returning an `Option<...>`? Or at least have one `get` variant that has this behavior?

If we talk about performance, the current `CFArrayIterator` use `CFArray::get` which means you have a bounds check for every call to `next` on the iterator. We actually have two since `CFArrayIterator::next` is also checking the bounds itself.

Is the array not guaranteed to be immutable? If so, the length of the array can be queried once and stored in the iterator for comparison without doing the ffi call to `CFArrayGetCount` every time. The iterator could then use a non-bounds checking (maybe unsafe?) version of the `get` method.

core-foundation/src/array.rs
jrmuizel: Rust, in theory, should let us use the existing reference in the array safely. Since its lifetime will be tied to the array we should be able to interact with those objects without needing to bump the retain count.

i.e. imagine if you have Vec<Rc<u32>>. You can iterate over them without having to bump the reference counts on all of the objects.

faern: Yes true. As long as the borrow can't outlive the array yes indeed!

How about you just `mem::forget` the `T` value in the `Drop` impl for `ItemRef`?

jrmuizel: fn drop takes &mut self so it won't let me just mem::forget() the T value

jrmuizel: But using ManuallyDrop let's me do it.

jrmuizel:
I implemented the bounds checking in #148

jrmuizel:
This is ready for review. It would be good to get this into the 0.5 release so we avoid breaking compat twice.

jdm:
@bors-servo: r+
Thanks for working on the details!

PR #152, opened 2018-01-26
Respect const/mut of system headers.

jdm:
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/152)


jdm:
Created in consultation with @jrmuizel.

@bors-servo r=jrmuizel

PR #153, opened 2018-01-26
Bump version to 0.5.1.

jdm:
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/153)


jdm:
@bors-servo r+
0.5.0 was yanked immediately, so no need to do another major increase for this.

PR #154, opened 2018-01-27
Add possibility to iterate untyped CFArray

faern:
Maybe it was intentional, or maybe it was forgotten. But it is not possible to iterate an untyped array (`CFArray<*const c_void>`) in the `0.5.1` release, it was in `0.4`.

This extra `FromVoid` impl on `*const c_void` fixes that.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/154)

### DIFF
--- a/core-foundation/src/array.rs
+++ b/core-foundation/src/array.rs
@@ -55,6 +55,12 @@ unsafe impl FromVoid for u32 {
     }
 }
 
+unsafe impl FromVoid for *const c_void {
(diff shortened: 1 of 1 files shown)


faern:
I ran into problems when upgrading rust-font-loader to use the `0.5` branch of core-foundation. It tries to iterate and map on an untyped array.

jrmuizel:
This looks good to me.

PR #156, opened 2018-02-01
Reduce usage of libc crate.

waywardmonkeys:
Switch from using `libc` types to `::std::os::raw`. This is a breaking
change due to the differences in how `c_void` is defined.

`core-foundation-sys` no longer needs the `libc` dependency.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/156)


waywardmonkeys:
Updated (and force pushed) with merge conflicts addressed.

PR #158, opened 2018-02-01
BUG: Add test showing how downcasting to CFArray is unsound

faern:
This PR is not intended to me merged directly. It adds a test that causes undefined behavior without using `unsafe {}`. So it's more to show the problem and discuss a solution.

The problem here is that if you have a `CFArray<X>` you can cast it up to a `CFType` and then down to a `CFArray<Y>` for any `Y: TCFType`. It never checks that the elements in the array are of type `Y` or not. At this point `CFArray::iter` will happily give out `&Y` references, while in fact the pointers backing those objects are pointing to `XRef` instances. So doing anything with the `&Y` will likely cause invalid memory access.

How do we solve this?

Ping @jrmuizel

---
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/158)

### DIFF
--- a/core-foundation/src/base.rs
+++ b/core-foundation/src/base.rs
@@ -64,6 +64,21 @@ impl CFType {
     /// assert!(cf_type.downcast::<CFBoolean>().is_none());
     /// ```
     ///
+    /// ```compile_fail
+    /// # use core_foundation::array::CFArray;
+    /// # use core_foundation::base::TCFType;
+    /// # use core_foundation::boolean::CFBoolean;
+    /// # use core_foundation::string::CFString;
+    /// #
+    /// let boolean_array = CFArray::from_CFTypes(&[CFBoolean::true_value()]).into_CFType();
+    ///
+    /// // This downcast is not allowed and causes compiler error, since it would cause undefined
+    /// // behavior to access the elements of the array as a CFString:
+    /// let invalid_string_array = boolean_array
+    ///     .downcast_into::<CFArray<CFString>>()
+    ///     .unwrap();
+    /// ```
+    ///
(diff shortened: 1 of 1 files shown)


faern:
I pushed another test, showing how this problem originates at `TCFType::instance_of`. Where
```rust
CFArray<X>::instance_of::<CFArray<Y>>()
```
Does return true in all cases.
I don't think the `instance_of` problem can be used to cause undefined behavior though. But it's still a strange situation.

jrmuizel:
Casting a CFArray to a particular CFArray<X> needs to be unsafe. I'm not yet sure how fix downcast.

jrmuizel:
If we add a trait that's ConcreteCFType that non-generic CFTypes implement. 

We can make downcast become
```rust
    pub fn downcast<T: ConcreteCFType>(&self) -> Option<T> {
```

and then we'd have
impl ConcreteCFType for CFArray<void*>
but not for CFArray<T>

jrmuizel:
I've put up a fix in #159

jdm:
We could potentially turn this into a doctest that verifies that it doesn't compile?

faern:
@jdm How do you create a test that passes if it does not compile?

jrmuizel:
See https://doc.rust-lang.org/beta/rustdoc/documentation-tests.html

```
/// ```compile_fail
/// let x = 5;
/// x += 2; // shouldn't compile!
/// ```
```

faern:
Updated the test to be a `compile_fail` doctest. Also fixed the trait bound on `downcast_into` to prohibit unsound downcasts there as well.

faern:
The `downcast_into<T: ConcreteCFType>` change here is breaking. But it's just as important for the security as the equivalent on `downcast` that was performed earlier.

PR #159, opened 2018-02-01
Add a ConcreteCFType trait

jrmuizel:
We'll use this to distinguish between generic and concrete CFTypes.
This lets us prevent downcast()ing to arbitrary generic types.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/159)


faern:
I have not tested it yet, but it looks like a reasonable solution. :+1:

PR #186, opened 2018-02-02
Unify the impl macros

jrmuizel:
Using some more macro magic we can avoid having to duplicate
our impl macro implementation for generic types. This will help
when we want a version for CFDictionary with two type parameters.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/186)


jdm:
@bors-servo r+
😰

PR #195, opened 2018-02-07
Add convertRectToBacking to NSView

jrmuizel:
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/195)


jdm:
What about https://github.com/servo/core-foundation-rs/blob/f86a5fb6810c3208d4a8c28df207432d23381ee5/cocoa/src/appkit.rs#L2789-L2791 which is already in the NSScreen type like https://developer.apple.com/documentation/appkit/nsscreen/1388389-convertrecttobacking?language=objc?

jrmuizel:
This is for https://developer.apple.com/documentation/appkit/nsview/1483648-convertrecttobacking. They are not the same method.

Issue #201, opened 2018-02-26
Implement a builder API for font variations


jdm:
Rather than having consumers pass in a stringly-typed dictionary for `new_from_variations`, it would be nice to provide a strong builder API that encapsulates that process.

PR #202, opened 2018-02-26
Only allow CFNumber font variations.

jdm:
Until #201 this seems like a worthwhile improvement.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/202)

### DIFF
--- a/core-text/Cargo.toml
+++ b/core-text/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "core-text"
-version = "9.1.2"
+version = "10.0.0"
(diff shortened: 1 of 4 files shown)


core-text/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "core-text"
-version = "9.1.2"
+version = "9.2.0"
```
jrmuizel: This needs to bump the major version.

jrmuizel:
@jdm do you mind finishing this up?

PR #208, opened 2018-05-02
Add NSViewFullScreenModeOptionKey keys

edwin0cheng:
This PR added NSViewFullScreenModeOptionKey values described in [here](https://developer.apple.com/documentation/appkit/nsviewfullscreenmodeoptionkey?language=objc)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/208)


edwin0cheng:
@vbo Should i placed these keys in that place, or should i add another extern block ?

PR #217, opened 2018-05-08
Update gleam to 0.5

jrmuizel:


kvark:
(a breaking change, but we just bumped the version and haven't published it, so it's fine)
@bors-servo r+

PR #219, opened 2018-05-28
Additional functions for `CGImage` and `CGDataProvider`

application-developer-DA:
1. Adds a possibility to construct a CGImage from a given provider. I don't know if it's the most optimal way to implement it and if it is safe, but it would be nice to wrap this function somehow, that was the first attempt to do it.
2. Adds some rendring intent constants.
3. Adds a function which allows to create a data provider from a slice. This is highly unsafe, that's why I marked it with unsafe. And I'm not sure if it is the best way to do this. Perhaps the one can store something like `Option<slice>` in the data provider, but in this case the provider will get some lifetime parameter (and we have to update the foreign type definition). I'm not sure that having slice parameter and a lifetime would make it 100% safe though, because when we're passing a data provider ref to some FFI, they could theoretically retain a data provider (or the image which uses the data provider), expecting that the data inside the data provider lives long enough. Probably the only "really safe" way to manage the data provider data is the one which is currently available (via `Arc<Vec<u8>>`), but this limitation of having `Arc<Vec<u8>>` may be very ineffective in some sort of appliations, so for instance in my case I had to add the additional "constructor" to allow creation of unsafe data provider and I guarantee the safety in my code by some other means.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/219)

### DIFF
--- a/core-graphics/src/base.rs
+++ b/core-graphics/src/base.rs
@@ -42,6 +42,12 @@ pub const kCGBitmapByteOrder32Little: u32 = (2 << 12);
 pub const kCGBitmapByteOrder16Big: u32 = (3 << 12);
 pub const kCGBitmapByteOrder32Big: u32 = (4 << 12);
 
+pub const kCGRenderingIntentDefault: u32 = 0;
+pub const kCGRenderingIntentAbsoluteColorimetric: u32 = 1;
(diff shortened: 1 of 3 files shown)


jdm:
@bors-servo r+
Thanks! Your reasoning about the safety of the new API makes sense.

PR #224, opened 2018-07-11
make core_graphics public for other project such as metal-rs to pass…

Michael-Lfx:
make core_graphics public for other project such as metal-rs to pass iOS compilation

### DIFF
--- a/cocoa/src/lib.rs
+++ b/cocoa/src/lib.rs
@@ -16,7 +16,7 @@ extern crate block;
 #[macro_use]
 extern crate bitflags;
 extern crate libc;
-extern crate core_graphics;
+pub extern crate core_graphics;
 #[macro_use]
 extern crate objc;
 


jrmuizel:
Why is this needed for iOS compilation?

Michael-Lfx:
Because [metal-rs](https://github.com/gfx-rs/metal-rs) use `NSSize` as return value. Compilation for macOS is good. For iOS, there is no `NSSize` definition in `Foundation`, instead we could use `CGSize` which defined in `Core Graphics` Framework. Here is **metal-rs** code snippet.

```
pub fn drawable_size(&self) -> NSSize {
    unsafe {
        msg_send![self, drawableSize]
    }
}

pub fn set_drawable_size(&self, size: NSSize) {
    unsafe {
        msg_send![self, setDrawableSize:size]
    }
}
```

jrmuizel:
Shouldn't metal-rs be using the type from the core graphics crate directly?

PR #226, opened 2018-07-16
Use new objc class macro

grovesNL:
This PR removes the `class` function in favor of the [new `class` macro exposed by objc](https://github.com/SSheldon/rust-objc/issues/65) which caches access by class name. Note that this macro should eventually (through procedural macros) be able to link directly to the class.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/226)


jdm:
@bors-servo r+
Thanks!

PR #248, opened 2018-08-14
Add support for image cropping

application-developer-DA:
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/248)

### DIFF
--- a/core-graphics/src/image.rs
+++ b/core-graphics/src/image.rs
@@ -5,6 +5,7 @@ use core_foundation::base::{CFRetain, CFTypeID};
 use core_foundation::data::CFData;
 use color_space::CGColorSpace;
 use data_provider::{CGDataProviderRef, CGDataProvider};
+use geometry::CGRect;
 use libc::size_t;
 use foreign_types::{ForeignType, ForeignTypeRef};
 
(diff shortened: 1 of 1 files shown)


core-graphics/src/image.rs
```diff
@@ -121,6 +122,10 @@ impl CGImageRef {
         };
         data_provider.copy_data()
     }
+
+    pub fn cropped(&self, rect: CGRect) -> CGImage {
+        unsafe { CGImage::from_ptr(CGImageCreateWithImageInRect(self.as_ptr(), rect)) }
```
jdm: This can return null if the rect defines an area that is not in the image: https://developer.apple.com/documentation/coregraphics/1454683-cgimagecreatewithimageinrect .

core-graphics/src/image.rs
```diff
@@ -121,6 +122,17 @@ impl CGImageRef {
         };
         data_provider.copy_data()
     }
+
+    /// Returns a cropped image. If the `rect` specifies a rectangle which lies outside of the
+    /// image bounds, the `None` is returned.
+    pub fn cropped(&self, rect: CGRect) -> Option<CGImage> {
```
jrmuizel: It might be good to include the fact that this uses CGImageCreateWithImageInRect and I'm pretty sure that makes it so that the underlying storage is shared between the two images.

PR #250, opened 2018-08-20
Allow most CGContext methods to be usable from a immutable reference and expose several more methods

mehcode:
If this is too much to submit at once, please tell me. This is what I've got left to send in and I'm unsure where to break it up more.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/250)

### DIFF
--- a/core-graphics/src/context.rs
+++ b/core-graphics/src/context.rs
@@ -9,9 +9,11 @@
 
 use base::CGFloat;
 use color_space::CGColorSpace;
-use core_foundation::base::{CFRelease, CFRetain, CFTypeID};
+use core_foundation::base::{ToVoid, CFRelease, CFRetain, CFTypeID};
 use font::{CGFont, CGGlyph};
 use geometry::CGPoint;
+use color::CGColor;
+use path::CGPathRef;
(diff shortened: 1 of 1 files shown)


jrmuizel:
Shouldn't these methods be taking a &mut? They are very much mutating the underlying object.

mehcode:
I don't disagree with you. I don't think `foreign_type! {` supports generating a `CGContextRefMut` structure to use though. Several clearly mutating methods that exist already took `&self` so I was copying what was there. Properly adding `_RefMut` and `&mut self` sounds like a breaking change and a large endeavor.

---

As a data point, my use case here is:

My use case here is:

```rust
let context: *mut Object = msg_send![class!(NSGraphicsContext), currentContext];
let context: *mut c_void = msg_send![context, CGContext];

let context = CGContextRef::from_ptr(context as *mut _);

context.save();
// [...]
context.restore();
```

Before this PR I was using `CGContext::wrap_under_` and `mem::forget`.

jrmuizel:
Ok. Let's take this.
@bors-servo r+

PR #254, opened 2018-09-13
enable u32 for CFNumber

sibsibsib:
kCFNumberIntType is needed in some iokit APIs. 

Please let me know if you think anything is missing.


core-foundation/src/number.rs
```diff
@@ -25,6 +25,25 @@ impl_CFTypeDescription!(CFNumber);
...
+            let ok = CFNumberGetValue(self.0, kCFNumberSInt32Type, mem::transmute(&mut value));
+            if ok { Some(value) } else { None }
+        }
+    }
+
+    #[inline]
+    pub fn to_u32(&self) -> Option<u32> {
+        unsafe {
+            let mut value: u32 = 0;
+            let ok = CFNumberGetValue(self.0, kCFNumberIntType, mem::transmute(&mut value));
```
jdm: I think you want kCFNumberSInt32Type and cast the resulting i32 to u32. Otherwise I'm worried that the resulting number might be different sizes on 32 vs. 64 bit architectures.

core-foundation-sys/src/number.rs
```diff
@@ -27,7 +27,7 @@ pub static kCFNumberFloat32Type:   CFNumberType = 5;
 pub static kCFNumberFloat64Type:   CFNumberType = 6;
 // static kCFNumberCharType:      CFNumberType = 7;
 // static kCFNumberShortType:     CFNumberType = 8;
-// static kCFNumberIntType:       CFNumberType = 9;
+pub static kCFNumberIntType:       CFNumberType = 9;
```
jdm: This change can be reverted with my suggestion.

core-foundation/src/number.rs
```diff
@@ -53,6 +72,21 @@ impl CFNumber {
     }
 }
 
+impl From<u32> for CFNumber {
+    #[inline]
+    fn from(value: u32) -> Self {
+        unsafe {
+            let number_ref = CFNumberCreate(
+                kCFAllocatorDefault,
+                kCFNumberIntType,
```
jdm: kCFNumberSInt32Type

jrmuizel:
CoreFoundation doesn't support a u32 type. Why do you need to add it?

sibsibsib:
I did some digging and found I am able to use the SInt32Type for the APIs I'm working with. As such, I've removed the u32-related code and left the to_i32 shortcut. Thanks for the feedback!

PR #259, opened 2018-09-27
Add initial bindings to Core Animation.

pcwalton:
Bound types include `CALayer`, `CATransaction`, and `CATransform3D`.

r? @jdm

---
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/259)

### DIFF
--- a/cocoa/src/quartzcore.rs
+++ b/cocoa/src/quartzcore.rs
@@ -1,0 +1,1587 @@
+// Copyright 2018 The Servo Project Developers. See the COPYRIGHT
+// file at the top-level directory of this distribution.
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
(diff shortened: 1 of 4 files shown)


jdm:
@jrmuizel Do you know things about CoreAnimation? I know nothing.

pcwalton:
@mstange has been reviewing my Core Animation related code in Gecko, FWIW.

mstange:
This CoreAnimation-specific parts looks straightfoward to me. What I have trouble with are the generic Objective C parts, and @jrmuizel is a better reviewer for those. For example the string conversion business and the use of `wrap_under_get_rule`, I'm not familiar with them.

cocoa/src/quartzcore.rs
```diff
@@ -0,0 +1,1589 @@
...
+    #[inline]
+    pub fn remove_from_superlayer(&self) {
+        unsafe {
+            msg_send![self.id(), removeFromSuperlayer]
+        }
+    }
+
+    /// This is only a snapshot, unlike the corresponding Cocoa API.
+    #[inline]
+    pub fn sublayers(&self) -> Vec<CALayer> {
```
jrmuizel: Why not just return a CFArray<CALayer>?

pcwalton: Good point, made that change.

PR #265, opened 2018-10-22
Converting between dictionary types

faern:
I'm trying to use the `CFDictionary` and `CFMutableDictionary` types more extensively. But run into problems. They are not very compatible with each other. This crate for example does not expose any clean way of using `CFDictionaryCreateMutableCopy` to create a `CFMutableDictionary` from any `CFDictionary`. I added a `From` implementation for this.

It was not possible to, in some safe/simple way, cast a `CFDictionary<K, V>` into the untyped `CFDictionary<*const c_void, *const c_void>` something I added `to_untyped` to be able to do.

When it comes to downcasting, it was not possible to downcast a `CFType` into a dictionary, since they don't implement `ConcreteCFType`. Something they probably should, for the case where the keys and values are just void pointers. `CFArray` does exactly this.

Lastly, what I needed the most was to be able to convert a `CFMutableDictionary` into a `CFDictionary`. This allows me to borrow an immutable `CFDictionary` via the `ItemRef` type basically. It's possible to work around this and actually have owned instances of a `CFDictionary` and a `CFMutableDictionary` pointing to the same actual dictionary by doing this:
```rust
let mut_dict = CFMutableDictionary::new();
let mut_dict2 = mut_dict.clone();
let dict = mut_dict.as_immutable();
// Here one can mutate the dict via `mut_dict2` while `dict` is still in scope/alive.
```
But since we already allow cloning a `CFMutableDictionary` to have two mutable references to the same dictionary I guess someone decided this should be viewed as safe. Both `CFDictionary` and `CFMutableDictionary` does not implement `Send` so we should not end up in any actual thread inconsistencies here.

If the above is indeed considered safe, I could get rid of the `ItemRef` juggling and just make it `fn as_immutable(&self) -> CFDictionary` directly maybe? Would simplify usage.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/265)

### DIFF
--- a/core-foundation/src/dictionary.rs
+++ b/core-foundation/src/dictionary.rs
@@ -177,10 +177,10 @@ impl<K, V> CFMutableDictionary<K, V> {
         unsafe { CFMutableDictionary::wrap_under_create_rule(reference) }
     }
 
-    /// Borrow self as an immutable dictionary.
+    /// Returns a `CFDictionary` pointing to the same underlying dictionary as this mutable one.
     #[inline]
-    pub fn as_immutable(&self) -> ItemRef<CFDictionary<K, V>> {
-        unsafe { CFDictionary::from_void(self.0 as _) }
+    pub fn to_immutable(&self) -> CFDictionary<K, V> {
+        unsafe { CFDictionary::wrap_under_get_rule(self.0) }
     }
 
     // Immutable interface
@@ -385,9 +385,9 @@ pub mod test {
         mut_dict.add(&CFString::from_static_string("Bar"), &CFBoolean::false_value());
         assert_eq!(mut_dict.retain_count(), 1);
 
-        let dict = mut_dict.as_immutable();
-        assert_eq!(mut_dict.retain_count(), 1);
-        assert_eq!(dict.retain_count(), 1);
+        let dict = mut_dict.to_immutable();
+        assert_eq!(mut_dict.retain_count(), 2);
+        assert_eq!(dict.retain_count(), 2);
         assert_eq!(*dict.get(&CFString::from_static_string("Bar")), CFBoolean::false_value());
 
         mem::drop(dict);
--- a/core-foundation/src/dictionary.rs
+++ b/core-foundation/src/dictionary.rs
@@ -235,6 +235,17 @@ impl<K, V> CFMutableDictionary<K, V> {
     }
 }
 
+impl<'a, K, V> From<&'a CFDictionary<K, V>> for CFMutableDictionary<K, V> {
+    /// Creates a new mutable dictionary with the key-value pairs from another dictionary.
+    /// The capacity of the new mutable dictionary is not limited.
+    fn from(dict: &'a CFDictionary<K, V>) -> Self {
+        unsafe {
+            let mut_dict_ref = CFDictionaryCreateMutableCopy(kCFAllocatorDefault, 0, dict.0);
+            TCFType::wrap_under_create_rule(mut_dict_ref)
+        }
+    }
+}
+
 
 #[cfg(test)]
 pub mod test {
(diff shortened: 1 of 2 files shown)


core-foundation/src/propertylist.rs
```diff
@@ -86,8 +86,9 @@ pub trait CFPropertyListSubClass: TCFType {
 
 impl CFPropertyListSubClass for ::data::CFData {}
 impl CFPropertyListSubClass for ::string::CFString {}
-impl CFPropertyListSubClass for ::array::CFArray {}
-impl CFPropertyListSubClass for ::dictionary::CFDictionary {}
+impl<T> CFPropertyListSubClass for ::array::CFArray<T> {}
+impl<K, V> CFPropertyListSubClass for ::dictionary::CFDictionary<K, V> {}
+impl<K, V> CFPropertyListSubClass for ::dictionary::CFMutableDictionary<K, V> {}
```
faern: I'm not completely sure this is correct or safe, but I think it is. The docs say that a mutable dictionary can be viewed as an immutable one, from my understanding. But not the other way around. So looking at a mutable dictionary as a propertylist should just be an extension of viewing it as an immutable dictionary then?

core-foundation/src/dictionary.rs
```diff
@@ -146,6 +165,26 @@ impl<K, V> CFMutableDictionary<K, V> {
...
+    #[inline]
+    pub fn into_untyped(self) -> CFMutableDictionary {
+        let reference = self.0;
+        mem::forget(self);
+        unsafe { CFMutableDictionary::wrap_under_create_rule(reference) }
+    }
+
+    /// Borrow self as an immutable dictionary.
+    #[inline]
+    pub fn as_immutable(&self) -> ItemRef<CFDictionary<K, V>> {
```
faern: As I wrote in the initial description above, trying to limit this usage so the mutable dictionary acts as borrowed (thus immutable) while the returned value exists might be overkill/false claims. Since it's so easy to go around it anyway. Anyone can just clone their mutable dictionary and make one of them immutable and still have a mutable version.

With the added addition of
```rust
 impl<K, V> CFPropertyListSubClass for ::dictionary::CFMutableDictionary<K, V> {} 
```
below (Which we have not yet debated if it's correct or not), they could also just do
```rust
let mut_dict = CFMutableDictionary::new();
let dict = mut_dict.to_CFPropertyList().downcast::<CFDictionary>().unwrap();
```
and that way also have an owned immutable dictionary be live at the same time as a mutable version pointing to the same underlying dictionary.

core-foundation/src/dictionary.rs
```diff
@@ -146,6 +165,26 @@ impl<K, V> CFMutableDictionary<K, V> {
...
+    pub fn into_untyped(self) -> CFMutableDictionary {
+        let reference = self.0;
+        mem::forget(self);
+        unsafe { CFMutableDictionary::wrap_under_create_rule(reference) }
+    }
+
+    /// Returns a `CFDictionary` pointing to the same underlying dictionary as this mutable one.
+    #[inline]
+    pub fn to_immutable(&self) -> CFDictionary<K, V> {
+        unsafe { CFDictionary::wrap_under_get_rule(self.0) }
```
faern: Since it was so easy to work around the borrowing here anyway I decided to revert that in the last commit. If we allow cloning of `CFMutableDictionary` I don't see why we would not allow this.

core-foundation/src/propertylist.rs
```diff
@@ -86,8 +86,8 @@ pub trait CFPropertyListSubClass: TCFType {
 
 impl CFPropertyListSubClass for ::data::CFData {}
 impl CFPropertyListSubClass for ::string::CFString {}
-impl CFPropertyListSubClass for ::array::CFArray {}
-impl CFPropertyListSubClass for ::dictionary::CFDictionary {}
+impl<T> CFPropertyListSubClass for ::array::CFArray<T> {}
+impl<K, V> CFPropertyListSubClass for ::dictionary::CFDictionary<K, V> {}
```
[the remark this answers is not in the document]
faern: How stupid of me.. This of course applies to for example `impl<T> CFPropertyListSubClass for ::array::CFArray<T> {}` as well. That would allow downcasting to a typed array directly, which would not be safe. I removed the generics on these `CFPropertyListSubClass` implementations again..

PR #275, opened 2018-11-09
add NSURL

acheronfail:
This PR adds in the trait for `NSURL` (requested in #183).

---
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/275)

### DIFF
--- a/cocoa/src/foundation.rs
+++ b/cocoa/src/foundation.rs
@@ -701,21 +701,26 @@ impl NSRunLoop for id {
     }
 }
 
-pub enum NSURLBookmarkCreationOptions {
-    NSURLBookmarkCreationPreferFileIDResolution = (1 << 8),
-    NSURLBookmarkCreationMinimalBookmark = (1 << 9),
-    NSURLBookmarkCreationSuitableForBookmarkFile = (1 << 10),
-    NSURLBookmarkCreationWithSecurityScope = (1 << 11),
-    NSURLBookmarkCreationSecurityScopeAllowOnlyReadAccess = (1 << 12),
-}
-
-pub enum NSURLBookmarkResolutionOptions {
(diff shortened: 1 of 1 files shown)


cocoa/src/foundation.rs
```diff
@@ -701,6 +701,336 @@ impl NSRunLoop for id {
...
 
+pub enum NSURLBookmarkCreationOptions {
+    NSURLBookmarkCreationPreferFileIDResolution = (1 << 8),
+    NSURLBookmarkCreationMinimalBookmark = (1 << 9),
+    NSURLBookmarkCreationSuitableForBookmarkFile = (1 << 10),
+    NSURLBookmarkCreationWithSecurityScope = (1 << 11),
+    NSURLBookmarkCreationSecurityScopeAllowOnlyReadAccess = (1 << 12),
+}
+
+pub enum NSURLBookmarkResolutionOptions {
```
acheronfail: Is the the correct way to define `NSURLBookmarkResolutionOptions` and `NSURLBookmarkCreationOptions`?

They're `NSUInteger` types, I'm not sure if this is the right way to define them...

For reference:

* https://developer.apple.com/documentation/foundation/nsurlbookmarkresolutionoptions?language=objc
* https://developer.apple.com/documentation/foundation/nsurlbookmarkcreationoptions?language=objc

jdm: It would be better to define these as bitflags; this would make it easier to combine them per https://developer.apple.com/documentation/foundation/nsurlbookmarkresolutionoptions?language=objc

acheronfail: Sorry! I must have missed this notification.

They're defined as `bitflags!` now, so we should be all good. :smile:

Issue #277, opened 2018-11-20
Fails to build on architectures where char is unsigned


infinity0:
See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=906612 and https://buildd.debian.org/status/package.php?p=rust-core-foundation

infinity0:
This includes for example aarch64 (arm v8 64 bit)

PR #279, opened 2018-12-07
Use libc c_char instead of i8 for libc strlen() call

silwol:
This patch allows build errors like https://buildd.debian.org/status/fetch.php?pkg=rust-core-foundation&arch=arm64&ver=0.6.1-1&stamp=1533301991&raw=0 to go through on architectures that use u8 as c_char.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/279)

### DIFF
--- a/core-foundation/src/url.rs
+++ b/core-foundation/src/url.rs
@@ -20,7 +20,7 @@ use std::ptr;
 use std::path::{Path, PathBuf};
 use std::mem;
 
-use libc::{strlen, PATH_MAX};
+use libc::{c_char, strlen, PATH_MAX};
 
 #[cfg(unix)]
 use std::os::unix::ffi::OsStrExt;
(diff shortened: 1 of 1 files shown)


nox:
I'm confused why this is built at all on Debian, given it's a macOS-specific crate.

nox:
Sure it causes no harm, but what package are you talking about? What is the crate that you are building on Debian which depends on this one? This code won't work elsewhere than on macOS.

infinity0:
> But what is the purpose of making a package that can't work because it is missing system APIs?

because cargo#5896.

How is this even relevant? core-foundation-rs has a bug, you are justifying not fixing it *because you think something else is strange*???

infinity0:
This PR would close #277 BTW.

nox:
> you are justifying not fixing

I never said anything about not fixing it, I'm just being curious about why this was even discovered by Debian, thanks for linking to that Cargo issue.

PR #292, opened 2019-02-01
Add CGImageDestination

jaysonsantos:
Hey there, this is an initial work to make CGImages saveable and it was tested with the code below.
I didn't see much tests around, should I add them with my PR?
I also put some TODOs on the code because those were questions that I could not answer.
Let me know if this is useful and I will keep with the work.

```rust
use core_graphics::display::{
    kCGNullWindowID, kCGWindowImageDefault, kCGWindowListOptionAll, CGDisplay, CGPoint, CGRect,
    CGSize,
};

use core_graphics::image::CGImageDestination;

use core_foundation::string::CFString;
use core_foundation::url::CFURL;

fn main() {
    let position = CGPoint::new(0.0, 0.0);
    let size = CGSize::new(1000.0, 1000.0);
    let bounds = CGRect::new(&position, &size);

    let screenshot = CGDisplay::screenshot(
        bounds,
        kCGWindowListOptionAll,
        kCGNullWindowID,
        kCGWindowImageDefault,
    )
    .unwrap();
    let destination = CGImageDestination::with_url(
        CFURL::from_path("screenshot.bmp", false).unwrap(),
        CFString::new("com.microsoft.bmp"),
        1,
        None,
    );
    destination.add_image(screenshot, None);
    assert!(destination.finalize());
}
```

### DIFF
--- a/core-graphics/src/image.rs
+++ b/core-graphics/src/image.rs
@@ -1,8 +1,11 @@
 use std::ptr;
 
 use base::CGFloat;
-use core_foundation::base::{CFRetain, CFTypeID};
+use core_foundation::base::{CFRetain, CFTypeID, CFRelease, TCFType};
 use core_foundation::data::CFData;
+use core_foundation::dictionary::{CFDictionary, CFDictionaryRef};
+use core_foundation::url::{CFURL, CFURLRef};
+use core_foundation::string::{CFString, CFStringRef};
 use color_space::CGColorSpace;
 use data_provider::{CGDataProviderRef, CGDataProvider};
 use geometry::CGRect;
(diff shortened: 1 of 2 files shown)


core-graphics/src/image.rs
```diff
@@ -163,3 +219,19 @@ extern {
     //fn CGImageGetAlphaInfo(image: ::sys::CGImageRef) -> CGImageAlphaInfo;
     //fn CGImageCreateCopyWithColorSpace(image: ::sys::CGImageRef, space: ::sys::CGColorSpaceRef) -> ::sys::CGImageRef
 }
+
+#[link(name = "ImageIO", kind = "framework")]
+extern "C" {
```
jaysonsantos: Does this mean it should be in a subproject called image-io?

kornelski: Yes, it'd be better to make it a separate crate, so that users of the crate can control which frameworks they depend on.

arbaaz:
What's the status on this?

kornelski:
The issue of linking to a new framework remains: https://github.com/servo/core-foundation-rs/pull/292#discussion_r253186185

This makes projects depend on ImageOI.framework in addition to CoreFoundation.framework. This is rather unexpected.

jrmuizel:
Yeah, this should go in a separate image-io crate

arbaaz:
Got it.

PR #297, opened 2019-03-14
Begin documenting core-foundation macros.

waywardmonkeys:
I started doing this after talking through some issues with @raphlinus. Clearly more to do, but would like to get this in sooner rather than later.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/297)


jdm:
@bors-servo r+
Thanks for writing this documentation!

PR #310, opened 2019-04-08
Change NSPasteboard sendData to setData to match Apple's API.

### LINKED ISSUE #309: Should NSPasteboard's sendData_forType() be setData_forType() instead?
I wanted to use NSPasteboard's ```setData(_:forType:)``` method, shown in [Apple's docs](https://developer.apple.com/documentation/appkit/nspasteboard/1531214-setdata).  Looking in ```cocoa::appkit::NSPasteboard```, I found ```sendData_forType``` ("send" instead of "set"), which is not in Apple's docs.  But I tried it anyway, and sure enough I got this error:
```
    -[NSPasteboard sendData:forType:]: unrecognized selector sent to instance 0x7fb47360dca0
```
The code in appkit.rs calls this:
```
    msg_send![self, sendData:data forType:dataType]
```
So I tried calling that directly in my code to reproduce the problem (it did), and then changed "send" to "set", and it worked perfectly.

I'm not a cocoa/objc expert, so I might be misunderstanding something.  But if it is there by mistake, I can submit a PR (and look for any similar issues).

Note that there is a "setData" in NSPasteboardItem.  But this is for NSPasteboard.

Assuming it's incorrect, this code in appkit.rs just needs the "send"s changed to "set"s.

```rust
pub trait NSPasteboard: Sized {
    // --snip--
    unsafe fn sendData_forType(self, data: id, dataType: id) -> BOOL;
    // --snip--
}

impl NSPasteboard for id {
    // --snip--
    unsafe fn sendData_forType(self, data: id, dataType: id) -> BOOL {
        msg_send![self, sendData:data forType:dataType]
    }
    // --snip--
}
```

ecpost:
Fixes #309

---
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/310)

### DIFF
--- a/cocoa/src/appkit.rs
+++ b/cocoa/src/appkit.rs
@@ -508,7 +508,7 @@ pub trait NSPasteboard: Sized {
 
     unsafe fn clearContents(self) -> NSInteger;
     unsafe fn writeObjects(self, objects: id) -> BOOL;
-    unsafe fn sendData_forType(self, data: id, dataType: id) -> BOOL;
+    unsafe fn setData_forType(self, data: id, dataType: id) -> BOOL;
     unsafe fn setPropertyList_forType(self, plist: id, dataType: id) -> BOOL;
     unsafe fn setString_forType(self, string: id, dataType: id) -> BOOL;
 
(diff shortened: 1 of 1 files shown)


jdm:
@bors-servo r+
Thanks!

PR #312, opened 2019-04-28
add barebones NSSpellChecker bindings

euclio:
Also adds `substringWithRange` to `NSString`.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/312)

### DIFF
--- a/cocoa/src/appkit.rs
+++ b/cocoa/src/appkit.rs
@@ -12,7 +12,7 @@
 use base::{id, BOOL, SEL};
 use block::Block;
 use foundation::{NSInteger, NSUInteger, NSTimeInterval,
-                 NSPoint, NSSize, NSRect, NSRectEdge};
+                 NSPoint, NSSize, NSRect, NSRange, NSRectEdge};
 use libc;
 
 pub use core_graphics::base::CGFloat;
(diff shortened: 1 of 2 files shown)


cocoa/src/appkit.rs
```diff
@@ -4017,6 +4017,66 @@ impl NSToolbar for id {
...
+    }
+
+    unsafe fn checkSpellingOfString_startingAt_language_wrap_inSpellDocumentWithTag_wordCount(
+        self,
+        stringToCheck: id,
+        startingOffset: NSInteger,
+        language: id,
+        wrapFlag: BOOL,
+        tag: NSInteger,
+        wordCount: *mut NSInteger) -> NSRange {
```
jdm: Let's return `(NSRange, NSInteger)` instead of accepting a mutable argument here.

euclio: Done.

cocoa/src/appkit.rs
```diff
@@ -4017,6 +4018,66 @@ impl NSToolbar for id {
...
+    }
+
+    unsafe fn checkSpellingOfString_startingAt_language_wrap_inSpellDocumentWithTag_wordCount(
+        self,
+        stringToCheck: id,
+        startingOffset: NSInteger,
+        language: id,
+        wrapFlag: BOOL,
+        tag: NSInteger) -> (NSRange, NSInteger) {
+        let wordCount = ptr::null_mut();
```
jdm: This won't do the right thing. We should initialize `let mut wordCount = 0;`, pass a mutable reference as the argument, and return the value of `wordCount`.

euclio: Whoops, fixed!

Issue #205, opened 2018-04-17
Higher-level interface to create delegates


vbo:
It seems currently the only way to create e.g. an NSWindowDelegate is through `objc::declare::ClassDecl` interface:
```rust
let superclass = Class::get("NSObject").unwrap();
let mut decl = ClassDecl::new("MyWindowDelegate", superclass).unwrap();
decl.add_method ...
```
Should this crate expose `trait NSWindowDelegate` and make the creation of such delegate more type-safe?

vbo:
True, this is not about Core Foundation, but it is about Cocoa, which is [a part](https://github.com/servo/core-foundation-rs/blob/master/cocoa/src/appkit.rs) of this crate.

This crate defines `trait NSWindow` for one of the core Cocoa interfaces. The interface provides `setDelegate` method to set `NSWindowDelegate` implementation which controls window behavior. Now currently the binding for this method looks like `setDelegate_(self, delegate: id)` so I can put any object there. I can create this object via `objc` crate, but I was wondering if this is too low-level and if we want to provide type-safe API in this crate to create such a delegate object.

nox:
Oh I see.

Litarvan:
Any update ? I was also using setDelegate and didn't find any other way than the one mentionned by @vbo , may I create a pull request to implement this ?

jdm:
A higher level interface sounds like a useful improvement, yes.

Litarvan:
I tried something in #314

PR #314, opened 2019-05-02
delegate! macro, a higher-level way to create Cocoa delegates (e.g. NSWindowDelegate)

Litarvan:
As @vbo mentioned in #205 the only way to create cocoa delegates such as NSWindowDelegate is to use the objc crate and it's a bit hard.

I thought a macro for this would be useful.

## Example with NSWindowDelegate

Without macro :
```rust
let mut decl = ClassDecl::new("MyWindowDelegate", class!(NSObject)).unwrap();

decl.add_ivar::<id>("window");
decl.add_method(sel!(windowWillEnterFullScreen:), on_enter_fullscreen as extern fn(&Object, Sel, id));

let cl = decl.register();

let delegate: id = msg_send![cl, alloc];
(*delegate).set_ivar("window", window);

window.setDelegate_(delegate);
```

With macro :
```rust
window.setDelegate_(delegate!("MyWindowDelegate", {
    window: id = window,
    (windowWillEnterFullScreen:) => on_enter_fullscreen as extern fn(&Object, Sel, id)
}));
```

---
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/314)

### DIFF
--- a/cocoa/src/lib.rs
+++ b/cocoa/src/lib.rs
@@ -28,3 +28,5 @@ pub mod base;
 pub mod foundation;
 #[cfg(target_os = "macos")]
 pub mod quartzcore;
+#[macro_use]
+mod macros;
--- a/cocoa/src/macros.rs
+++ b/cocoa/src/macros.rs
@@ -16,22 +16,26 @@
 /// #[macro_use] extern crate objc;
 ///
 /// use cocoa::appkit::NSWindow;
-/// use cocoa::base::nil;
+/// use cocoa::base::{id, nil};
 ///
-/// # unsafe fn main() {
-/// let my_window: id = NSWindow::alloc(nil);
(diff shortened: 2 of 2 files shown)


vbo:
This is a good change. Although at the time I was thinking more about type-safety of the interface than the amount of typing required to set things up.

Litarvan:
In fact, i tried looking for such thing, but i saw that NSWindowDelegate is just a protocol, which doesn't really mean something, so i thought making a macro like this for any delegate would be great

jdm:
Looks like a nice improvement!
@bors-servo r+

vbo:
Can we add a test/example code with this? Or maybe update one of the existing examples?

Litarvan:
I added an example in the doc comment, but if you want to add something feel free to do it @vbo

PR #318, opened 2019-06-14
Fix `CGDisplayMode` bindings

aleksijuvani:
There was some bizarre behaviour going on with `CGDisplayMode::all_display_modes()`, which I believe was resulting in objects being freed that weren't supposed to. In my project, all calls to `CGConfigureDisplayWithDisplayMode()` would crash after this function call, and any subsequent calls to `CGDisplayMode::all_display_modes()` would crash as well.

Familiarising myself with this codebase, I think the right solution here is to wrap the array elements with `TCFType::wrap_under_get_rule()` as indicated by the comment in the `CFArray::iter()` function. In order to do this, instead of the `foreign_type!()` macro we will have to use `impl_TCFType!()` for the `CGDisplayMode` type.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/318)

### DIFF
--- a/core-graphics/src/display.rs
+++ b/core-graphics/src/display.rs
@@ -11,13 +11,13 @@
 
 use libc;
 use std::ptr;
-use std::ops::Deref;
+use std::os::raw::c_void;
 
 pub use base::{CGError, boolean_t};
 pub use geometry::{CGRect, CGPoint, CGSize};
 
 use core_foundation::string::{CFString, CFStringRef};
-use core_foundation::base::{CFRetain, TCFType};
+use core_foundation::base::{TCFType, CFTypeID};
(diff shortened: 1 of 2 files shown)


aleksijuvani:
I'm not entirely convinced that this is correct either. I'm having a hard time wrapping my head around the implementation details of this crate, so I've worked around the issue by rolling my own bindings and using carefully placed `CGDisplayModeRetain` and `CGDisplayModeRelease` calls.

jdm:
Sorry for forgetting about this. Could you file an issue with sample code that reproduces the problem?

aleksijuvani:
I've filed issue #319 to track this.

PR #320, opened 2019-07-14
Fix use of deprecated libc::(u)int{16,32}_t

poiru:
This fixes warnings like:

>use of deprecated item 'libc::int32_t': Use i32 instead.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/320)


poiru:
cc @jdm

jdm:
@bors-servo r+
Thanks!

PR #322, opened 2019-07-14
Fix NSColor argument order: r b<->g

felixrabe:
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/322)


jdm:
@bors-servo r+
Weird that these methods were named this way originally!

PR #337, opened 2019-10-11
Add CFDataGetBytes to CFData bindings

wezm:
Just add a missing binding that I needed.

This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/337)


wezm:
I rebased this but looks like it needs a nudge to rebuild.

PR #340, opened 2019-10-16
Fixes so the compiler can infer msg_send! return types

SSheldon:
Currently, due to a quirk in Rust's type inference interacting with the structure of the `msg_send!` macro, a `()` return type will be inferred when the compiler cannot otherwise determine the return type. This behavior is expected to change, and in the future could resolve to a `!` return type, which results in undefined behavior.

Linting has previously been added for this in rust-lang/rust#39216, but it did not catch these cases due to SSheldon/rust-objc#62. An upcoming version of objc will be fixed to stop hiding these errors, at which point they will become compile errors.

This change fixes these errors and allows cocoa to compile with the fixed version of objc.

### DIFF
--- a/cocoa/src/appkit.rs
+++ b/cocoa/src/appkit.rs
@@ -549,7 +549,7 @@ pub trait NSPasteboard: Sized {
 
 impl NSPasteboard for id {
     unsafe fn releaseGlobally(self) {
-        msg_send![self, releaseGlobally];
+        msg_send![self, releaseGlobally]
     }
 
     unsafe fn clearContents(self) -> NSInteger {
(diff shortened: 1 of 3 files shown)


jdm:
@bors-servo r+
Thanks!

SSheldon:
Thanks @jdm!

Related: how do y'all feel about backported versions here? There's still crates relying on cocoa 0.18 that will need these fixes (primarily glutin and winit). glutin and winit have already moved on to developing alpha versions of their next releases. So either we can backport a cocoa 0.18 fix here or they can backport fixes that update to cocoa 0.19 to get this fix.

If I can do something to make a cocoa 0.18.5, that means fewer PRs overall for me 😛

SSheldon:
As I'm exploring more of the dependency graph, I'm thinking the backport option would be more and more helpful 😛

If you're amenable to it, I've prepared a branch that backports this fix: https://github.com/SSheldon/core-foundation-rs/tree/backport_0.18.5

SSheldon:
@jdm @SimonSapin is there anyone I can talk to about a backport release? Sorry, not sure the right venue/way to bring this up. I think a backport for 0.18 would be really helpful; as described in https://github.com/rust-lang/rust/pull/65355#issuecomment-542974382, over a third of the downloads for cocoa are still the 0.18 release.

jdm:
A backport release makes sense. I'll try and figure out the logistics of how to do that based on your branch.

SSheldon:
@jdm if we want a bit more of a paper trail: if you create a new empty branch based off of 8ffba8a79adf0b6aaf162917127a2a556e24698c, I can open a PR to merge my backport changes into that. No worries if that seems unnecessary, I'm sure you've got it covered.

jdm:
@SSheldon Done: https://github.com/servo/core-foundation-rs/tree/cocoa-0.18

SSheldon:
Perfect, I opened #342, will continue any discussion there. Thanks much @jdm!

PR #341, opened 2019-10-16
Update cocoa version to 0.19.1

SSheldon:
New version needed so that the #340 fixes can be published.


SSheldon:
@jdm thank you! Who can take care of tagging this and publishing it to crates, is that something I also ask you about?

jdm:
Tagged and published!

Issue #339, opened 2019-10-15
Release changes since March


wezm:
The last release was 0.6.4 on Mar 30, 2019. Would be great to get some of the recent changes into a release.

jdm:
Which crate are you talking about? We've definitely released some since then.

wezm:
I was referring to [core-foundation](https://crates.io/crates/core-foundation). It would be great to have a release of `core-text` with my recent changes too though.

zubivan:
Would be great to get a release of `core-graphics` as well.

jdm:
If people submit PRs with cargo.toml version updates for the crates that should be released, I'll merge and publish them.

zubivan:
@jdm bumped CG version to the next minor.

jdm:
Opened #346 for core-foundation, core-foundation-sys, and core-text.

atouchet:
#346 was merged. Can these updates be published to crates.io?

jdm:
I've just published the updates from #349. Is that enough?

atouchet:
I guess so. It would have been nice to include #228 (or better yet an update to 0.8) if you were going to do a breaking change.

atouchet:
There have been new releases of core-foundation, core-text, and core-graphics. I think this should be fine to close now.

PR #345, opened 2019-10-26
Bump core graphics version to 0.18.0

zubivan:
As discussed in #339 bumping version of core graphics to the next minor.

### DIFF
--- a/cocoa/Cargo.toml
+++ b/cocoa/Cargo.toml
@@ -17,6 +17,6 @@ block = "0.1"
 bitflags = "1.0"
 libc = "0.2"
 core-foundation = { path = "../core-foundation", version = "0.6" }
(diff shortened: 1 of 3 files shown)


jdm:
Published.

zubivan:
@jdm thanks!

nox:
core-graphics is exposed in core-text's public API. This should have been a core-text breaking change. Please don't update crates without updating the others, now I have releases to yank.

jdm:
To be clear - nothing in this PR is incorrect, because core-text's version was not updated at all. I did a major version update in #346.

nox:
My complaint is that there was no paper trail that there were unpublished breaking changes in the tree, and thus core-text actually needed a breaking bump.

PR #349, opened 2019-11-12
Fix CFRunLoop types to avoid UB

wez:
A number of callback functions associated with the CFRunLoop types are
allowed to be NULL to evoke default behavior.

Prior to this commit the definition of those structs prevented passing
in null which meant resorting to transmute or other tricks to
effectively set null, but with recent compiler versions the compiler
emits `ud2` instructions and triggers a fault at runtime.

The correct resolution for this is to define these fields as `Option`al
callbacks and that is what this commit does.

I've used this approach successfully here in another project:
Refs: https://github.com/wez/wezterm/commit/398f333c32fb1ff4ccd9cc001a66ae22522eb60c

I've bumped up the version for the crate as part of this commit
because it effectively changes the API around these structs.


lqf96:
Since the definition of `CFRunLoopTimerContext` has changed, I think the test in `core-foundation/src/runloop.rs` should also be updated... It's a simple update though, just replace `mem::zeroed()` with `None`...

core-foundation-sys/src/runloop.rs
```diff
@@ -50,25 +50,25 @@ pub const kCFRunLoopAllActivities: CFOptionFlags = 0x0FFFFFFF;
...
-    pub schedule: extern "C" fn (info: *const c_void, rl: CFRunLoopRef, mode: CFStringRef),
-    pub cancel: extern "C" fn (info: *const c_void, rl: CFRunLoopRef, mode: CFStringRef),
+    pub retain: Option<extern "C" fn (info: *const c_void) -> *const c_void>,
+    pub release: Option<extern "C" fn (info: *const c_void)>,
+    pub copyDescription: Option<extern "C" fn (info: *const c_void) -> CFStringRef>,
+    pub equal: Option<extern "C" fn (info1: *const c_void, info2: *const c_void) -> Boolean>,
+    pub hash: Option<extern "C" fn (info: *const c_void) -> CFHashCode>,
+    pub schedule: Option<extern "C" fn (info: *const c_void, rl: CFRunLoopRef, mode: CFStringRef)>,
+    pub cancel: Option<extern "C" fn (info: *const c_void, rl: CFRunLoopRef, mode: CFStringRef)>,
     pub perform: extern "C" fn (info: *const c_void),
```
jdm: What about `perform`?

core-foundation-sys/src/runloop.rs
```diff
@@ -50,25 +50,25 @@ pub const kCFRunLoopAllActivities: CFOptionFlags = 0x0FFFFFFF;
...
-    pub equal: extern "C" fn (info1: *const c_void, info2: *const c_void) -> Boolean,
-    pub hash: extern "C" fn (info: *const c_void) -> CFHashCode,
+    pub retain: Option<extern "C" fn (info: *const c_void) -> *const c_void>,
+    pub release: Option<extern "C" fn (info: *const c_void)>,
+    pub copyDescription: Option<extern "C" fn (info: *const c_void) -> CFStringRef>,
+    pub equal: Option<extern "C" fn (info1: *const c_void, info2: *const c_void) -> Boolean>,
+    pub hash: Option<extern "C" fn (info: *const c_void) -> CFHashCode>,
     // note that the following two fields are platform dependent in the C header, the ones here are for macOS
     pub getPort: extern "C" fn (info: *mut c_void) -> mach_port_t,
     pub perform: extern "C" fn (msg: *mut c_void, size: CFIndex, allocator: CFAllocatorRef, info: *mut c_void) -> *mut c_void,
```
jdm: What about getPort and perform?

io-surface/Cargo.toml
```diff
@@ -10,6 +10,6 @@ license = "MIT / Apache-2.0"
 [dependencies]
 libc = "0.2"
 gleam = "0.7"
-core-foundation = { path = "../core-foundation", version = "0.6" }
+core-foundation = { path = "../core-foundation", version = "0.7" }
```
jdm: This crate now needs to be bumped to 0.13.0.

core-text/Cargo.toml
```diff
@@ -15,5 +15,5 @@ mountainlion = []
 [dependencies]
 foreign-types = "0.3"
 libc = "0.2"
-core-foundation = { path = "../core-foundation", version = "0.6.2" }
+core-foundation = { path = "../core-foundation", version = "0.7" }
```
jdm: This crate now needs to be bumped to 15.0.0.

core-graphics/Cargo.toml
```diff
@@ -14,6 +14,6 @@ highsierra = []
 
 [dependencies]
 bitflags = "1.0"
-core-foundation = { path = "../core-foundation", version = "0.6" }
+core-foundation = { path = "../core-foundation", version = "0.7" }
```
jdm: This crate needs to be bumped to 0.19.0.

cocoa/Cargo.toml
```diff
@@ -16,7 +16,7 @@ crate-type = ["rlib"]
 block = "0.1"
 bitflags = "1.0"
 libc = "0.2"
-core-foundation = { path = "../core-foundation", version = "0.6" }
+core-foundation = { path = "../core-foundation", version = "0.7" }
```
jdm: This crate needs to be bumped to 0.20.0.

core-foundation-sys/src/runloop.rs
wez: perform cannot be null

core-foundation-sys/src/runloop.rs
wez: these cannot be null

PR #353, opened 2020-01-09
Add binding for CTRunGetStringIndices

maxbrunsfeld:
This PR adds a `CTRun::string_indices` method, which exposes CoreText's `CTRunGetStringIndices`/`CTRunGetStringIndicesPtr` API. The implementation is basically identical to `CTRun::glyphs` and `CTRun::positions`.

I added an assertion about the new method in the existing unit test.

### DIFF
--- a/core-graphics/src/event.rs
+++ b/core-graphics/src/event.rs
@@ -12,9 +12,9 @@ pub type CGEventField = u32;
 pub type CGKeyCode = u16;
 pub type CGScrollEventUnit = u32;
 
-/// Flags for events
-///
-/// [Ref](http://opensource.apple.com/source/IOHIDFamily/IOHIDFamily-700/IOHIDSystem/IOKit/hidsystem/IOLLEvent.h)
+// Flags for events
+//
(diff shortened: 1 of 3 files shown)


maxbrunsfeld:
I added an unrelated commit that just fixes a few warnings in the codebase, because I wanted to depend on this branch in another project. If you'd prefer these fixes *not* be included in this PR, I can remove that commit.

core-graphics/src/event.rs
```diff
@@ -12,9 +12,9 @@ pub type CGEventField = u32;
 pub type CGKeyCode = u16;
 pub type CGScrollEventUnit = u32;
 
-/// Flags for events
-///
-/// [Ref](http://opensource.apple.com/source/IOHIDFamily/IOHIDFamily-700/IOHIDSystem/IOKit/hidsystem/IOLLEvent.h)
+// Flags for events
```
jdm: Could you revert this change and move the doc comment to the struct inside the macro?

jdm:
@bors-servo r+
thanks!

maxbrunsfeld:
⚡️Thanks for the quick response!

PR #356, opened 2020-02-13
Add new methods to NSView

not-a-seagull:
https://developer.apple.com/documentation/appkit/nsview/

I have added setFrameOrigin and setFrameSize to the `NSView` trait and its `id` implementation.


jdm:
@bors-servo r+
Thanks!

PR #363, opened 2020-03-21
Assert when creating a CFType from null

### LINKED ISSUE #361: Missing null checks
In the `core-foundation` crate, there appears to be a *lot* of functions missing NULL checks in the allocation, potentially causing segfaults when trying to over-allocate (instead of panicking or returning a Result). Examples include [dictionary creation](https://github.com/servo/core-foundation-rs/blob/master/core-foundation/src/dictionary.rs#L46), [array creation](https://github.com/servo/core-foundation-rs/blob/master/core-foundation/src/array.rs#L68) or [string creation](https://github.com/servo/core-foundation-rs/blob/master/core-foundation/src/string.rs#L111). Those functions can all return NULL according to Apple's documentation (e.g. [see return value doc here](https://developer.apple.com/documentation/corefoundation/1516791-cfdictionarycreatemutable?language=objc)

It would seem wise to add some railguards against this somewhere - perhaps `wrap_under_create_rule` should have an explicit NULL-check, causing a panic if it is encountered?

roblabla:
What's the reason behind IOSurface not using `impl_TCFType`? It feels dangerous not to use it.

Fixes #361

### DIFF
--- a/core-foundation/src/base.rs
+++ b/core-foundation/src/base.rs
@@ -243,6 +243,7 @@ impl TCFType for CFType {
 
     #[inline]
     unsafe fn wrap_under_get_rule(reference: CFTypeRef) -> CFType {
+        assert!(!reference.is_null(), "Attempted to create a NULL object.");
         let reference: CFTypeRef = CFRetain(reference);
         TCFType::wrap_under_create_rule(reference)
     }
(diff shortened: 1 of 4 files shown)


jdm:
I don't know! If it can be, it probably should.

PR #368, opened 2020-04-13
Add CGContextSetShadow and CGContextSetShadowWithColor to CGContext

brunogouveia:


core-graphics/src/context.rs
```diff
@@ -705,5 +717,8 @@ extern {
 
     fn CGContextDrawLinearGradient(c: ::sys::CGContextRef, gradient: ::sys::CGGradientRef, startPoint: CGPoint, endPoint: CGPoint, options: CGGradientDrawingOptions);
     fn CGContextDrawRadialGradient(c: ::sys::CGContextRef, gradient: ::sys::CGGradientRef,  startCenter: CGPoint, startRadius: CGFloat, endCenter:CGPoint, endRadius:CGFloat, options: CGGradientDrawingOptions);
+
+    fn CGContextSetShadow(c: ::sys::CGContextRef, offset: CGSize, blur: CGFloat);
+    fn CGContextSetShadowWithColor(c: ::sys::CGContextRef, offset: CGSize, blur: CGFloat, color: *const c_void);
```
jdm, line 722: Why the void*? https://developer.apple.com/documentation/coregraphics/1455205-cgcontextsetshadowwithcolor says it should be a CGColorRef.

brunogouveia: I think I declared in the same way `CGContextSetFillColorWithColor` was declared, but we can change the color type to `::sys::CGColorRef` and import TCFType trait.

Btw, for future reference, why some types were defined with `foreign_type` whilst others were defined with `declare_TCFType`?

jdm: I think we started using foreign_type on an ad-hoc basis at some point, and only in some repositories before merging them into this one? In any case, if it builds, foreign_type is usually preferred now.

jdm:
@bors-servo r+
Thanks!

PR #370, opened 2020-04-17
Change CFMutableAttributedString::set_attribute argument type

brunogouveia:
Type of the argument `value` can be a reference as opposed to requiring the value's ownership.


jrmuizel:
FWIW, this changed the api so it should've bumped the version number. See https://github.com/linebender/piet/pull/220

jdm:
I'm correcting that in https://github.com/servo/core-foundation-rs/pull/397.

PR #374, opened 2020-04-29
feat: added bindings to NSUserDefaults

casperstorm:
I have added bindings for [`NSUserDefaults`](https://developer.apple.com/documentation/foundation/nsuserdefaults). It is used in macOS for storing state, but it may also be used read/write state which is already set by the system. One example could be that an application would like to overwrite `CGFontRenderingFontSmoothingDisabled` for enabling sub-pixel antialiasing on > 10.14.

Let me know if I can improve anything.


jdm:
@bors-servo r+
Thanks!

PR #375, opened 2020-05-06
feat: extended NSProgressInfo with more functions

casperstorm:
I have extended `NSProgressInfo` with two new functions:

* [`operatingSystemVersion`](https://developer.apple.com/documentation/foundation/nsprocessinfo/1410906-operatingsystemversion?language=objc) which returns a `NSOperatingSystemVersion` where you can get `majorVersion`, `minorVersion` and `patchVersion` from.

* [`isOperatingSystemAtLeastVersion`](https://developer.apple.com/documentation/foundation/nsprocessinfo/1414876-isoperatingsystematleastversion?language=objc) which takes a `NSOperatingSystemVersion` struct, and returns a `bool`.

Let me know if I missed something.

### DIFF
--- a/cocoa/src/foundation.rs
+++ b/cocoa/src/foundation.rs
@@ -203,17 +203,48 @@ impl NSAutoreleasePool for id {
     }
 }
 
+
+#[repr(C)]
+#[derive(Copy, Clone)]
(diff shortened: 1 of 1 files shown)


jdm:
@bors-servo r+
Thanks!

PR #378, opened 2020-05-08
Make CFString comparable

### LINKED ISSUE #104: PartialEq and Eq for CFString.
I wish `CFString` had implementations for `Eq` and `PartialEq` using `CFStringCompare`. `Ord` and `PartialOrd` could also be done using the same function! :)

kornelski:
Fixes #104

The `PartialEq` has to be on `CFString`, not `CFStringRef` due to orphan rules.

### DIFF
--- a/core-foundation/src/string.rs
+++ b/core-foundation/src/string.rs
@@ -141,6 +141,53 @@ impl CFString {
     }
 }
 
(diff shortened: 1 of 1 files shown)


jdm:
@bors-servo r+
Thanks!

PR #379, opened 2020-05-08
Test 64-bit NSNumbers

kornelski:
on macOS 10.15.5 Beta the property list tests fails, because dictionary serializes a 32-bit NSNumber, but property list deserializes it to a 64-bit NSNumber. Apparently, they make dictionaries not equal, even though numeric value is the same!

Here's a workaround that forces the NSNumber to be 64-bit.


jdm:
@bors-servo r+
Thanks!

PR #380, opened 2020-05-09
Add CTFramesetter::suggest_frame_size_with_constraints

cmyr:
I'm not sure if there's a policy of trying to match the documented API exactly? It feels nice to not take pointers for return values, but I'm happy to do it either way.

### DIFF
--- a/core-text/src/framesetter.rs
+++ b/core-text/src/framesetter.rs
@@ -7,13 +7,15 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-use std::os::raw::c_void;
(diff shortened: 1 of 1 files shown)


core-text/src/framesetter.rs
```diff
@@ -37,15 +39,43 @@ impl CTFramesetter {
...
+                null(),
+            );
 
             CTFrame::wrap_under_create_rule(ptr)
         }
     }
+
+    /// Suggest an appropriate frame size for displaying a text range.
+    ///
+    /// Returrns a tuple containing an appropriate size (that should be smaller
```
jdm, line 54: nit: Returns

PR #383, opened 2020-05-09
Add CTFrame::get_lines and related methods

cmyr:
This is a bit unsatisfactory; there doesn't seem to be a memory safe
way to get access to the lines without an intermediate allocation.

As a compromise, this adds API for also checking the number of lines
in the frame, and getting a single line safely.

### DIFF
--- a/core-text/src/frame.rs
+++ b/core-text/src/frame.rs
@@ -9,8 +9,10 @@
 
 use std::os::raw::c_void;
 use core_foundation::base::{CFTypeID, TCFType};
+use core_foundation::array::{CFArrayRef, CFArray};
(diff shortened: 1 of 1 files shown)


cmyr:
I appreciate this is taking a few more liberties with the API. I think the minimal version of this would be to return a vec of retained CTLines, and I'm happy to change the patch to only include that method, and the additional stuff can be discussed separately.

jdm:
Yeah, I'm uncomfortable with the ways this PR works around the underlying CTFrameGetLines API. I'd prefer to merge the changes that retain all of the lines.

cmyr:
sounds good!

cmyr:
okay, should be uncontroversial

PR #359, opened 2020-03-18
CFSet: Add missing methods in -sys

roblabla:
Adds a handful of missing methods from the CFSet family.

PR #387, opened 2020-05-17
Bump cocoa to 0.20.1

chrisduerr:
The new operating system version and user defaults APIs have been recently added by @casperstorm, so I'd like to propose releasing a new version with those features included.

There hasn't been a version in a bit, so I think it should be fine to create a new non-breaking release, as long as there are no other things currently in the pipeline. It seems like #359 and #363 are both unrelated to cocoa, but maybe #355 can be easily merged before this?


jdm:
Thanks for the reminder! Since #355 includes a version bump, I'll go ahead and close this PR and publish after it merges.

PR #389, opened 2020-06-22
Build cocoa docs for docs.rs on macOS

kraai:
I'm not sure that this will work, but I don't know how to test it locally.  https://docs.rs/about#metadata says that this will cause the package to be built on macOS.


kraai:
Would it be possible to upload a new version of cocoa to crates.io to determine whether this fixes the documentation on docs.rs?

jdm:
Done.

kraai:
Thanks, looks great: https://docs.rs/cocoa/0.20.2/cocoa/.

PR #390, opened 2020-06-23
Fix docs.rs for core-text and io-surface

kraai:
See #389.


jdm:
I just pushed commits that did this. Thanks anyways!

kraai:
Ah, great!  Thanks!

PR #391, opened 2020-06-26
Remove gleam dependency from io-surface.

### LINKED ISSUE #371: Remove gleam dependency from io-surface
It's only used for 5 GL constants and 1 type (GLenum). We can write out the constant values directly and use the GLenum from [cgl-rs](https://github.com/servo/cgl-rs/blob/master/src/cgl.rs#L17).

jdm:
Fixes #371.

### DIFF
--- a/io-surface/Cargo.toml
+++ b/io-surface/Cargo.toml
@@ -3,7 +3,7 @@ name = "io-surface"
 description = "Bindings to IO Surface for macOS"
 homepage = "https://github.com/servo/core-foundation-rs"
 repository = "https://github.com/servo/core-foundation-rs"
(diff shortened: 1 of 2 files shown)


io-surface/Cargo.toml
```diff
@@ -3,7 +3,7 @@ name = "io-surface"
 description = "Bindings to IO Surface for macOS"
 homepage = "https://github.com/servo/core-foundation-rs"
 repository = "https://github.com/servo/core-foundation-rs"
-version = "0.13.1"
+version = "0.13.2"
```
atouchet, line 6: The version number will need to be updated.

PR #357, opened 2020-02-21
Fix for link error - undefined CFMutableAttributedStringGetTypeID

vbadanov:
## Intention
Fix for https://github.com/rust-lang/rust/issues/63047 and https://github.com/intellij-rust/intellij-rust/issues/4605

## How to reproduce
The following program fails to compile:

```rust
use core_foundation::attributed_string::{CFAttributedString, CFMutableAttributedString};
use core_foundation::base::TCFType;

fn main() {
    dbg!(<CFAttributedString as TCFType>::type_id());
    dbg!(<CFMutableAttributedString as TCFType>::type_id());
}
```

```
error: linking with `cc` failed: exit code: 1
  |
[...]

PR #394, opened 2020-06-28
Bump core-foundation(-sys) to 0.7.1 for release.

anp:
Hi! I'm hoping to be able to pull https://github.com/servo/core-foundation-rs/pull/357 from crates.io and based on other issue threads here it seems that a PR with version bumps is the way to request a release.

I haven't used my local mac recently so I'm relying on CI for now, although if needed I can spin that up and test it locally. LMK?

I'm not actually sure whether the changes to core-foundation since 2019/11 qualify for a patch version bump or a minor one. Thoughts?


jdm:
I've double checked the changes to core-foundation and core-foundation-sys and a minor version change looks good for both. Thanks!

jdm:
Published.

jdm:
FYI: we ended up having to do a major version bump after all and yank the new minor releases.

anp:
Appreciate the heads up, sorry for the hassle!

PR #403, opened 2020-07-17
Separate out some new crates

jrmuizel:
This spits out core-graphics-types and cocoa-foundation with the goal of reducing the amount of interdependence between the crates with goal of reducing churn when things need to be updated.

cocoa-foundation should be enough for metal-rs and gfx-metal-backend.

Originally, I had wanted core-graphics-types to be enough for core-text but it's use of CGContext prevents that.


jdm:
```
error[E0463]: can't find crate for `cocoa`

 --> cocoa-foundation/tests/foundation.rs:4:1

  |

4 | extern crate cocoa;

  | ^^^^^^^^^^^^^^^^^^^ can't find crate

error: aborting due to previous error
```

jrmuizel:
This passes tests now.

Cargo.toml
```diff
@@ -1,2 +1,2 @@
 [workspace]
-members = ["core-foundation", "core-foundation-sys", "core-graphics", "core-text", "cocoa", "io-surface"]
+members = ["core-foundation", "core-foundation-sys", "core-graphics-types", "core-graphics", "core-text", "cocoa", "io-surface"]
```
jdm, line 2: This is missing cocoa-foundation.

jrmuizel: Fixed.

PR #406, opened 2020-08-26
Fix docs.rs build

kornelski:
docs.rs fails to build `core-graphics`. This adds metadata to let it know to build on an Apple platform.


jdm:
@bors-servo r+
Thanks!

PR #408, opened 2020-09-04
Add API for NSView's layerContentsPlacement property

tronical:


tronical:
Apologies, I had a silly typo in the setter, where the selector for the getter was chosen instead of the correct setter 🤦

PR #409, opened 2020-09-05
Add strikethrough attribute name.

smmalis37:
I don't have a mac to test this on, I'm going purely based on this doc https://developer.apple.com/documentation/uikit/nsstrikethroughstyleattributename.


smmalis37:
I think i've misread the docs, this might not work.

smmalis37:
Ignore this, I don't know how macs work lol.

PR #411, opened 2020-10-04
add `CFNumberGetType` to determine specific number type

sassman:
this PR adds support for `CFNumberGetType` that is needed if there is a `CFNumberRef` and it is not clear what sort of number it is.

Now it would be possible to do such a thing:

```rust
let type_id = unsafe { CFNumberGetType(value) };
match type_id {
    si64 if si64 == kCFNumberSInt64Type => {},
    si32 if si32 == kCFNumberSInt32Type => {},
    _ => {},
}
```


sassman:
Do you need any further things to land this PR?

PR #412, opened 2020-10-04
feat(CFNumber): change `CFNumberType` from static to const, to improve match expression ergonomics

sassman:
this PR changes the `CFNumberType` values like `kCFNumberSInt64Type` from `static` to `const`.

That way it is possible to simplify match expressions like this:

before:
```rust
let type_id = unsafe { CFNumberGetType(value) };
match type_id {
    si64 if si64 == kCFNumberSInt64Type => {},
    si32 if si32 == kCFNumberSInt32Type => {},
    _ => {},
}
```

after:
```rust
let type_id = unsafe { CFNumberGetType(value) };
match type_id {
    kCFNumberSInt64Type => {},
    kCFNumberSInt32Type => {},
    _ => {},
}
```

Note: function call `CFNumberGetType(value)` is implemented in PR #411

### DIFF
--- a/core-foundation-sys/src/number.rs
+++ b/core-foundation-sys/src/number.rs
@@ -19,23 +19,23 @@ pub type CFBooleanRef = *const __CFBoolean;
 pub type CFNumberType = u32;
 
 // members of enum CFNumberType
-// static kCFNumberSInt8Type:     CFNumberType = 1;
-// static kCFNumberSInt16Type:    CFNumberType = 2;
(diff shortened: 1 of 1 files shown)


sassman:
Do you need any further things to land this PR?

jrmuizel:
Is there any way that this change can break existing code? i.e. do we need to bump the version for it?

sassman:
@jrmuizel it should not break anything. But I added a test to illustrate the former and the current usage to make sure it's handled in the build.

PR #413, opened 2020-10-04
feat(CFBoolean): add function `CFBooleanGetValue` to retrieve a value of a `CFBooleanRef`

sassman:
This PR adds the function `CFBooleanGetValue` that gets the bool value by a `CFBooleanRef`


sassman:
@jdm Thanks for merging. Any idea when this will be releases as new version? 
A bit of background: I just finished a [new tool](https://github.com/sassman/t-rec-rs) that relies on the patches #413 #412 #411 getting released.

PR #415, opened 2020-10-13
Fix the definition of CFAllocatorContext

jrmuizel:
The function pointers can be null so should be wrapped in Option.

This struct was basically unusable before so I think we should
be fine making this change while only bumping the patch version.
I was unable to find any uses of it on github.


jrmuizel:
@jdm can you do a new release of core-foundation-sys?

PR #421, opened 2020-10-26
Test that copying system fonts doesn't change the url

jrmuizel:


core-text/src/font.rs
```diff
@@ -660,3 +661,13 @@ fn copy_font() {
     assert_eq!(font.family_name(), "Zapf Dingbats");
 }
 
+#[test]
+fn copy_system_font() {
+    let small = unsafe {
+        CTFont::wrap_under_create_rule(
+        CTFontCreateUIFontForLanguage(kCTFontSystemDetailFontType, 19., std::ptr::null())
```
jdm, line 668: nit: indentation.

core-text/src/font.rs
```diff
@@ -539,7 +539,8 @@ extern {
     fn CTFontCreateWithFontDescriptor(descriptor: CTFontDescriptorRef, size: CGFloat,
                                       matrix: *const CGAffineTransform) -> CTFontRef;
     //fn CTFontCreateWithFontDescriptorAndOptions
-    //fn CTFontCreateUIFontForLanguage
+    #[cfg(test)]
```
jdm, line 542: Why is this only for tests?

jdm (commented):
```
thread 'font::copy_system_font' panicked at 'assertion failed: `(left == right)`

  left: `Some(file:///System/Library/Fonts/SFNSDisplay-Regular.otf)`,

 right: `Some(file:///System/Library/Fonts/SFNSText-Regular.otf)`', core-text/src/font.rs:672:5

note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

test font::copy_system_font ... FAILED
```

core-text/src/font.rs
jrmuizel: I'm currently only using it tests and otherwise we get a warning about it being unused.

core-text/src/font.rs
jrmuizel: > ```
> thread 'font::copy_system_font' panicked at 'assertion failed: `(left == right)`
> 
>   left: `Some(file:///System/Library/Fonts/SFNSDisplay-Regular.otf)`,
> 
>  right: `Some(file:///System/Library/Fonts/SFNSText-Regular.otf)`', core-text/src/font.rs:672:5
> 
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> 
> test font::copy_system_font ... FAILED
> ```

This was intentional. I plan to use these tests to help document the semantics of these functions.

jrmuizel:
apologies for the garbage macos_version() function

PR #424, opened 2020-10-27
Make CTFontDescriptor Send and Sync and expose attributes()

jrmuizel:


core-text/src/font_descriptor.rs
```diff
@@ -194,6 +194,10 @@ declare_TCFType! {
 impl_TCFType!(CTFontDescriptor, CTFontDescriptorRef, CTFontDescriptorGetTypeID);
 impl_CFTypeDescription!(CTFontDescriptor);
 
+// "Font objects (CTFont, CTFontDescriptor, and associated objects) can be used
+//  simultaneously by multiple operations, work queues, or threads."
```
emilio, line 198: Can you link to the docs?

emilio (approved):

core-text/src/font_descriptor.rs
jrmuizel: Oh I missed this comment before I r+ it...

PR #425, opened 2020-10-28
Add github actions CI for macos-11.0

jrmuizel:


emilio:
(It won't block landing I think, because it probably isn't integrated with homu, but it's something)

PR #430, opened 2020-12-15
fix: new_from_descriptor lose font variations, such as font-weight

kaiwk:
If we create `CTFont` from font descriptor, we will lose font variations, and use `new_from_name` instead will fix it.

[CTFontDescriptor](https://developer.apple.com/documentation/coretext/ctfontdescriptor?language=objc) is essentially a dictionary of font attributes.


jrmuizel:
Can you add a test?

kaiwk:
@jrmuizel I have to add `CTFontManagerCreateFontDescriptorsFromData`, because it happens when load font from 'otc/ttc'.

I encounter the issue when using [font-kit](https://github.com/servo/font-kit), font-kit [manually unpack 'otc/ttc'](
https://github.com/servo/font-kit/blob/f0cb880b66310e8e43fb923b8d9416c2a2183997/src/loaders/core_text.rs#L826), I suppose it won't make any difference here.

kaiwk:
The test failed on xcode 7.3/9.2 and passed others.

core-text/src/font.rs
```diff
@@ -127,7 +129,21 @@ pub fn new_from_descriptor(desc: &CTFontDescriptor, pt_size: f64) -> CTFont {
 
 pub fn new_from_buffer(buffer: &[u8]) -> Result<CTFont, ()> {
     let ct_font_descriptor = create_font_descriptor(buffer)?;
-    Ok(new_from_descriptor(&ct_font_descriptor, 16.0))
+    Ok(new_from_name(&ct_font_descriptor.font_name(), 16.0)?)
```
jrmuizel, line 132: Won't this break if the font is not installed or if two different fonts use the same name? i.e. new_from_name won't find the font.

kaiwk: Well, looks like WebRender also create CTFont from `postscript name`, see: https://phabricator.services.mozilla.com/D93518 
>        // there's no way great way to go from a CGFont to a CTFontDescriptor
>        // so we use the postscript name. Ideally NativeFontHandle would
>        // just use a CTFontDescriptor
>        let name = native_font_handle.0.postscript_name();
>        let font = core_text::font_descriptor::new_from_postscript_name(&name);
>
>        self.ct_font_descs
>            .insert(*font_key, font);

Also, [`font_name()` returns `postscript name`](https://developer.apple.com/documentation/coretext/kctfontnameattribute?language=objc), and [`new_fomr_name` receives `postscript name`](https://developer.apple.com/documentation/coretext/1509153-ctfontcreatewithname?language=objc).

kaiwk:
I also add an api called `new_from_arc_buffer`, so maybe `font-kit` can benefit from it : )

core-text/src/font.rs
jrmuizel: WebRender only uses the postscript name for installed system fonts. We don't use the name for data fonts:
https://searchfox.org/mozilla-central/source/gfx/wr/webrender/src/platform/macos/font.rs#382

PR #454, opened 2021-04-09
Add bindings for NSOpenPanel

maxbrunsfeld:
This PR adds bindings for the main `NSOpenPanel` APIs, which let you present a modal for selecting files or directories. Usage:

```rust
let panel = NSOpenPanel::openPanel(nil);
panel.setCanChooseDirectories_(YES);
panel.setCanChooseFiles_(NO);
panel.setAllowsMultipleSelection_(YES);

let response = panel.runModal();
if response == NSModalResponse::NSModalResponseOk {
    let urls = panel.URLs();
    for i in 0..urls.count() {
        let url_ptr = urls.objectAtIndex(i).absoluteString().UTF8String();
        let url = CStr::from_ptr(url_ptr).to_string_lossy();
        eprintln!("You picked URL {}", url);
    }
}
```

PR #457, opened 2021-05-05
Add NSSavePanel binding

maxbrunsfeld:
This PR adds bindings to some basic `NSSavePanel` APIs. In Cocoa, `NSSavePanel` is a superclass of `NSOpenPanel`, which I added bindings for in #454.

Example:

```rust
let panel = NSSavePanel::savePanel(nil);
panel.setDirectoryURL(the_url);
let response = panel.runModal();
if response == NSModalResponse::NSModalResponseOk {
    let url_ptr = panel.URL().absoluteString().UTF8String();
    let url = CStr::from_ptr(url_ptr).to_string_lossy();
    eprintln!("You picked URL {}", url);
}

```

### DIFF
--- a/cocoa/src/appkit.rs
+++ b/cocoa/src/appkit.rs
@@ -1786,7 +1786,36 @@ pub enum NSModalResponse {
     NSModalResponseCancel = 0,
 }
 
-pub trait NSOpenPanel: Sized {
+pub trait NSSavePanel: Sized {
+    unsafe fn savePanel(_: Self) -> id {
+        msg_send![class!(NSSavePanel), savePanel]
+    }
+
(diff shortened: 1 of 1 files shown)


maxbrunsfeld:
The CI failure seems unrelated:

```
---- filedescriptor::test::test_consumed stdout ----
thread 'filedescriptor::test::test_consumed' panicked at 'assertion failed: `(left == right)`
  left: `0`,
 right: `-1`', core-foundation/src/filedescriptor.rs:129:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

On both `master` and this branch, I am able to reproduce it locally *occasionally*, by running the tests in a loop:

```sh
while cargo test; do echo ok; done
```

PR #458, opened 2021-05-11
foundation: add bundleIdentifier and resourcePath properties

nacho:
Adds the bundleIdentifier and resourcePath properties to the NSBundle
bindings


jdm:
@bors-servo r+
Thanks!

PR #460, opened 2021-08-11
Adds run_in_mode wrapper over CFRunLoopRunInMode

xanathar:
Adds a `run_in_mode` wrapper function which just makes `CFRunLoopRunInMode` a little more "rusty".

Also adds a `CFRunLoopRunResult` enum to type the return value.


jdm:
@bors-servo r+
Nice improvement!

xanathar:
Minor change: added some `#[derive]` to the result enum to simplify its handling.

PR #472, opened 2021-10-31
Prepare github actions CI for use by homu.

jrmuizel:


.github/workflows/rust.yml
```diff
@@ -2,21 +2,34 @@ name: Rust
 
 on:
   push:
-    branches: [ master ]
+    branches: ["**"]
```
jdm, line 5: I recommend making this "auto", otherwise the tests will be run post-merge on the master branch as well

jdm (approved):

.github/workflows/rust.yml
jrmuizel: I copied it from here: https://github.com/servo/font-kit/blob/master/.github/workflows/ci.yml#L4. I plan on making travis not run on master but figured it wasn't worth it for github actions and it can be nice to to have the test status show up in an expected place ("master")

PR #473, opened 2021-11-01
Don't run travis on master or 10.15

jrmuizel:
10.15 is now covered by Github actions and master is unneeded
because of auto


jrmuizel:
Hmm, it seems that excepting 'master' breaks CI on PRs. Presumably, because they use 'master' as a base?

jrmuizel:
Thanks to https://travis-ci.community/t/excluding-master-branch-but-allowing-prs-based-off-of-master/12341 I was able to figure this out.

mstange (approved):
@bors-servo r+

jrmuizel:
Looks like this is causing builds to run on `auto` twice.

jrmuizel:
It looks like maybe that was happening before.

jdm:
Yeah, that's a limitation - first homu resets auto to the current master, then it pushes the new merge commit. I'm not actually sure why it does that.

jrmuizel:
Do you know where the code that does that is?

jdm:
I suspect it comes from this code: https://github.com/servo/homu/blob/d6b918f45b4aadfb43ae4cba5b9b80a4f9c00638/homu/main.py#L624-L646

PR #477, opened 2021-11-16
Add ProcessInfo.systemUptime implementation

n-eq:
Hello,

Added [NSProcessInfo.systemUptime](https://developer.apple.com/documentation/foundation/processinfo/1414553-systemuptime) implementation, successfully tested on iOS.
Feel free to merge.

Best,

### DIFF
--- a/cocoa-foundation/src/foundation.rs
+++ b/cocoa-foundation/src/foundation.rs
@@ -242,7 +242,7 @@ impl NSProcessInfo for id {
         msg_send![self, processName]
     }
 
-    unsafe fn uptime(self) -> NSTimeInterval {
+    unsafe fn systemUptime(self) -> NSTimeInterval {
         msg_send![self, systemUptime]
     }
 
--- a/cocoa-foundation/src/foundation.rs
+++ b/cocoa-foundation/src/foundation.rs
@@ -229,9 +229,7 @@ pub trait NSProcessInfo: Sized {
         msg_send![class!(NSProcessInfo), processInfo]
     }
 
-    unsafe fn systemUptime(_: Self) -> NSTimeInterval {
-        msg_send![class!(NSProcessInfo), systemUptime]
-    }
+    unsafe fn systemUptime(_: Self) -> NSTimeInterval;
 
     unsafe fn processName(self) -> id;
     unsafe fn uptime(self) -> NSTimeInterval;
--- a/cocoa-foundation/src/foundation.rs
+++ b/cocoa-foundation/src/foundation.rs
@@ -229,7 +229,12 @@ pub trait NSProcessInfo: Sized {
         msg_send![class!(NSProcessInfo), processInfo]
     }
 
+    unsafe fn systemUptime(_: Self) -> NSTimeInterval {
+        msg_send![class!(NSProcessInfo), systemUptime]
+    }
+
     unsafe fn processName(self) -> id;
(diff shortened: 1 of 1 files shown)


cocoa-foundation/src/foundation.rs
```diff
@@ -229,7 +229,12 @@ pub trait NSProcessInfo: Sized {
         msg_send![class!(NSProcessInfo), processInfo]
     }
 
+    unsafe fn systemUptime(_: Self) -> NSTimeInterval {
+        msg_send![class!(NSProcessInfo), systemUptime]
```
jrmuizel, line 233: Why does this need a default implementation instead of just relying on the implementation below like the other functions? i.e. why isn't this implemented more like operatingSystemVersion etc.

n-eq: It doesn't need a default implementation. We can actually have the same implementation as `operatingSystemVersion` , good point.

cocoa-foundation/src/foundation.rs
```diff
@@ -229,7 +229,12 @@ pub trait NSProcessInfo: Sized {
...
+    }
```
n-eq, line 234:
```suggestion
    unsafe fn systemUptime(_: Self) -> NSTimeInterval;
```

cocoa-foundation/src/foundation.rs
```diff
@@ -229,7 +229,10 @@ pub trait NSProcessInfo: Sized {
         msg_send![class!(NSProcessInfo), processInfo]
     }
 
+    unsafe fn systemUptime(_: Self) -> NSTimeInterval;
+
```
n-eq, line 233:
```suggestion
    unsafe fn systemUptime(_: Self) -> NSTimeInterval;
```

cocoa-foundation/src/foundation.rs
```diff
@@ -239,6 +242,10 @@ impl NSProcessInfo for id {
         msg_send![self, processName]
     }
 
+    unsafe fn uptime(self) -> NSTimeInterval {
+        msg_send![self, systemUptime]
+    }
```
n-eq, line 247:
```suggestion
    unsafe fn systemUptime(self) -> NSTimeInterval {
        msg_send![self, systemUptime]
    }
```

cocoa-foundation/src/foundation.rs
```diff
@@ -229,7 +229,10 @@ pub trait NSProcessInfo: Sized {
         msg_send![class!(NSProcessInfo), processInfo]
     }
 
+    unsafe fn systemUptime(_: Self) -> NSTimeInterval;
+
     unsafe fn processName(self) -> id;
+    unsafe fn uptime(self) -> NSTimeInterval;
```
n-eq, line 235:
```suggestion
    unsafe fn systemUptime(_: Self) -> NSTimeInterval;
    unsafe fn processName(self) -> id;
```

cocoa-foundation/src/foundation.rs
```diff
@@ -229,6 +229,7 @@ pub trait NSProcessInfo: Sized {
         msg_send![class!(NSProcessInfo), processInfo]
     }
 
+    unsafe fn systemUptime(_: Self) -> NSTimeInterval;
```
n-eq, line 232:
```suggestion
    unsafe fn systemUptime(self) -> NSTimeInterval;
```

PR #494, opened 2022-01-21
core-foundation-rs: Upgrade to edition 2021

komi1230:
Today's latest Rust edition is 2021. 
This current module system is a bit older, and I upgraded this.


jrmuizel:
This bumps the minimum required Rust version so I'm not eager to take it without a compelling advantage. It also looks like it has a bunch of formatting changes mixed in.

PR #500, opened 2022-02-16
Add get name function to CTFont

ssssota:
## Overview

Now we can get family name, full name, postscript name and some names.
But I want to use other names. (e.g. Copyright notice, License etc...)
So, I made `get_string_by_name_key` methods from helper function.


jdm:
@bors-servo r+
Thanks!

PR #516, opened 2022-06-27
fix: add animate property to NSWindow:setFrame:

keiya01:
I added animate property to `NSWindow:setFrame:`.
If this behavior is intended, I'm sorry but please close this PR.


madsmtm (changes requested):
[`-[NSWindow setFrame:display:]`](https://developer.apple.com/documentation/appkit/nswindow/1419753-setframe?language=objc) and [`-[NSWindow setFrame:display:animate:]`](https://developer.apple.com/documentation/appkit/nswindow/1419519-setframe?language=objc) are different methods, so I think you should add `setFrame_display_animate_` instead of changing the existing `setFrame_display_`.

keiya01:
> [`-[NSWindow setFrame:display:]`](https://developer.apple.com/documentation/appkit/nswindow/1419753-setframe?language=objc) and [`-[NSWindow setFrame:display:animate:]`](https://developer.apple.com/documentation/appkit/nswindow/1419519-setframe?language=objc) are different methods, so I think you should add `setFrame_display_animate_` instead of changing the existing `setFrame_display_`.

I see. I fixed your review.

PR #517, opened 2022-06-27
Merge upstream changes from Warp's core-foundation-rs fork

alokedesai:
Hey servo folks, 
Sending a PR to merge [Warp](https://github.com/warpdotdev)'s changes to core-foundation-rs upstream. Note, these changes are purely additive, and shouldn't have an affect on any existing core-foundation-rs APIs:

## Changes
1. Adds `get_matrix` to `CTFont`,  matching [CTFontGetMatrix](https://developer.apple.com/documentation/coretext/1510407-ctfontgetmatrix?changes=_8).
2. Adds `get_typographic_bounds` to `CTRun`, matching [CTRunGetTypographicBounds](https://developer.apple.com/documentation/coretext/1510569-ctrungettypographicbounds)
3. Adds a patch for a panic in `extract_number_for_key` that occurs in Ventura. The underlying issue is an issue within the Core Framework library on Ventura--but this patch allows this library to not panic on Ventura if someone tries to access any of the [font traits](https://developer.apple.com/documentation/coretext/ctfontdescriptor/font_traits).

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -2,21 +2,34 @@ name: Rust
 
 on:
   push:
-    branches: [ master ]
+    branches: ["**"]
   pull_request:
-    branches: [ master ]
+    branches: ["**"]
 
 env:
   CARGO_TERM_COLOR: always
 
 jobs:
   build:
-
-    runs-on: macos-11.0
-
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [macos-11.0, macos-10.15]
     steps:
     - uses: actions/checkout@v2
     - name: Build
       run: cargo build --verbose
     - name: Run tests
       run: cargo test --verbose
+  build_result:
+    name: homu build finished
+    runs-on: ubuntu-latest
+    needs:
+      - "build"
+    steps:
+      - name: Mark the job as successful
+        run: exit 0
+        if: success()
+      - name: Mark the job as unsuccessful
+        run: exit 1
+        if: "!success()"
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 Cargo.lock
 target/
+.idea
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 os: osx
 language: rust
 rust: stable
+if: branch != master OR type != push
 matrix:
   include:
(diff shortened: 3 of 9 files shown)


jrmuizel:
Can you clean up the commit history and avoid the formatting changes?

jrmuizel:
Can you also give some example code or add a test case for reproducing the Ventura failure?

vorporeal:
In the name of finally getting (at least some of) these changes merged upstream, I've sent out https://github.com/servo/core-foundation-rs/pull/674 and https://github.com/servo/core-foundation-rs/pull/675 for the bits adding new functions.

In terms of the Ventura fix, I can't think of an easy way to write a test for it.  FWIW, this has been live in Warp (used by ~100k macOS users) for a year now with no indication of it causing any issues.  I can send out a separate PR for that as well, or we can keep it in our own fork; either works for us.

nicoburns:
> I can send out a separate PR for that as well, or we can keep it in our own fork; either works for us.

I definitely think we will want this fix upstream. A separate PR might make sense at this point seeing as it has been done for the other changes.

> In terms of the Ventura fix, I can't think of an easy way to write a test for it.

I think the best test for this would be a usage of the API that triggers the bug without the fix if you have a standalone reproduction (perhaps using a particular system font). But unless anyone else strongly objects, I think it would be better to land this without a test than not at all if you are unable to provide that. A link to the original bug in warp that lead to this fix would also potentially be helpful.

I'm sorry that progress on reviewing/merging this has been so slow. The Servo project is still getting back up to speed after a long hiatus and is still quite behind on maintenance of libraries. Hopefully this will improve over the next few months.

vorporeal:
Yeah we (@alokedesai or I) can put together a separate PR for the `extract_number_for_key` fix.

No need to apologize for latency here - there was an explicit request to clean up the commit history and provide some example/test code for the `extract_number_for_key` issue, and nobody from our side followed up. :)

jdm (approved):
I'm fine with these changes but it would be nice to clean up the commit history as previously requested.

vorporeal:
With the PR you just approved, we'll be able to close this out - I've already merged separate PRs for the other bits.

(Feel free to close this one out now and we can focus on getting #735 merged.)

jdm:
Cool, let's do that!

PR #126, opened 2017-11-11
Add CFDate, CFTimeZone and make them work with chrono

ratake:
This change is [Reviewable](https://reviewable.io/reviews/servo/core-foundation-rs/126)

PR #521, opened 2022-08-17
remove with-chrono feature

esheppa:
We are currently keen to use `core-foundation` in `chrono` to determine the system time zone name - via [`iana-time-zone`](https://lib.rs/crates/iana-time-zone) (and potentially timezone calculations later), however becuase `core-foundation` depends on `chrono`, `iana-time-zone` has to use `core-foundation-sys` instead. 

I'm proposing here that the `chrono` related code is removed from `core-foundation` and I'll create a separate PR to add it into `chrono`. I recognize this is a breaking change which is not ideal, however I've checked the top reverse dependencies from https://lib.rs/crates/core-foundation/rev and it seems like none of them are using the `with-chrono` feature. If this PR is impractical, I'm keen to explore if there are any other options available.

### DIFF
--- a/core-foundation/Cargo.toml
+++ b/core-foundation/Cargo.toml
@@ -15,13 +15,11 @@ version = "0.8.3"
 
 [dependencies]
 libc = "0.2"
-chrono = { version = "0.4", optional = true }
 uuid = { version = "0.5", optional = true }
 
 [features]
 mac_os_10_7_support = ["core-foundation-sys/mac_os_10_7_support"] # backwards compatibility
 mac_os_10_8_features = ["core-foundation-sys/mac_os_10_8_features"] # enables new features
-with-chrono = ["chrono"]
 with-uuid = ["uuid"]
 
 [package.metadata.docs.rs]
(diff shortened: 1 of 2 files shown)


esheppa:
I've created https://github.com/chronotope/chrono/pull/783 which brings all the functions from the `with-chrono` feature into `chrono`

djc:
(Seems like the chrono integration was added in https://github.com/servo/core-foundation-rs/pull/126 and never touched since.)

madsmtm (approved):
`cacao` was affected by this as well, see https://github.com/ryanmcgrath/cacao/pull/44.

Note that they didn't actually use the feature, was just a leftover.

esheppa:
There is further discussion at https://github.com/chronotope/chrono/pull/783 surrounding [potential users](https://github.com/chronotope/chrono/pull/783#issuecomment-1217985210) of the feature, and it seems to be all vendored versions of this crate

jrmuizel:
This seems reasonable. I'll see about other things we'd want to include in a major version bump.

djc:
@mrobinson thoughts on this? Would be nice to get this in.

mrobinson:
This seems reasonable to me as well. I guess this plus #549 could be good candidates for a major version bump.

djc:
New PR up at #666.

mrobinson:
Closing this as it's duplicated by #666.

PR #522, opened 2022-09-02
Trim dependencies

madsmtm:
Of particular note:
- `libc` is not needed in most cases, one can simply use `std::os::raw`
- `leaky-cow` seems like it existed to work around compiler limitations, these are no more

@jrmuizel I've included version bumps in here as well, would be nice if you could also do a new release after this

### DIFF
--- a/core-foundation/src/url.rs
+++ b/core-foundation/src/url.rs
@@ -18,8 +18,9 @@ use core_foundation_sys::base::{kCFAllocatorDefault, Boolean};
 use std::fmt;
 use std::ptr;
 use std::path::{Path, PathBuf};
+use std::os::raw::c_char;
 
-use libc::{c_char, strlen, PATH_MAX};
+use libc::{strlen, PATH_MAX};
 
 #[cfg(unix)]
 use std::os::unix::ffi::OsStrExt;
--- a/core-foundation/src/url.rs
+++ b/core-foundation/src/url.rs
@@ -14,18 +14,10 @@ pub use core_foundation_sys::url::*;
 use base::{TCFType, CFIndex};
 use string::{CFString};
 
-use core_foundation_sys::base::{kCFAllocatorDefault, Boolean};
+use core_foundation_sys::base::kCFAllocatorDefault;
 use std::fmt;
 use std::ptr;
-use std::path::{Path, PathBuf};
-use std::os::raw::c_char;
-
-use libc::{strlen, PATH_MAX};
-
-#[cfg(unix)]
-use std::os::unix::ffi::OsStrExt;
-#[cfg(unix)]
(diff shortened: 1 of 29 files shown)


madsmtm:
Ping @jdm? This would be really nice to have in `winit`, I'd be able to remove a lot of old cruft from that

core-foundation/src/url.rs
```diff
@@ -75,15 +74,20 @@ impl CFURL {
 
     #[cfg(unix)]
     pub fn to_path(&self) -> Option<PathBuf> {
+        // https://github.com/rust-lang/libc/blob/9d4e7802112e3d3186bdd0191a036281506c28b7/src/unix/linux_like/mod.rs#L1026
+        pub const PATH_MAX: c_int = 4096;
```
jrmuizel, line 78: PATH_MAX is 1024 on macOS

madsmtm: You're right, I didn't check this properly; have brought back `libc` for this, but have made it possible to disable under a feature flag instead.

madsmtm:
Tests are passing for this now, would you mind taking another look?

complexspaces:
FWIW, it would be nice if the changes here moving everything towards using `std` wasn't applied universally. Notably, crates like `core-foundation-sys` and other `-sys` crates that contain nothing but bindings. These didn't strictly change much in this PR, but I wanted to bring up the idea since this seemed like a good place.

You stated this goal in the PR description:
> libc is not needed in most cases, one can simply use std::os::raw

However, depending on the MSRV goals (if any) that these crates have, I would argue that using `libc` is the better option to make more of these available in `no_std` contexts to anyone who may want that. [I personally found it surprising](https://github.com/1Password/sys-locale/commit/6717e5662a93d820ac4b18a1a31716dccd304049) this weekend to find that `core-foundation-sys` used `std` given how little it does.

A lot of the C FFI types just moved to `core` in recent Rust versions that obsoletes the above, but it is probably too new for these crates given how widely used they are.

jrmuizel:
> However, depending on the MSRV goals (if any) that these crates have, I would argue that using `libc` is the better option to make more of these available in `no_std` contexts to anyone who may want that. [I personally found it surprising](https://github.com/1Password/sys-locale/commit/6717e5662a93d820ac4b18a1a31716dccd304049) this weekend to find that `core-foundation-sys` used `std` given how little it does.

@complexspaces What's the use case for using core-foundation-sys in a no-std environment?

complexspaces:
> What's the use case for using core-foundation-sys in a no-std environment?

@jrmuizel Binary sizes, primarily.

madsmtm:
I think most of this has been done in other PRs, so I'll leave it for now

PR #525, opened 2022-10-06
Add CGContextResetClip

IjzerenHein:
Adds the `CGContextResetClip` function as `context.reset_clip`.

From the `CGContext.h` file:

```
/* Reset the current clip of `c' to the default value. */

CG_EXTERN void CGContextResetClip(CGContextRef c);
```

### DIFF
--- a/core-graphics/src/context.rs
+++ b/core-graphics/src/context.rs
@@ -389,6 +389,12 @@ impl CGContextRef {
         }
     }
 
+    pub fn reset_clip(&self) {
+        unsafe {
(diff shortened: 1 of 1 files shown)


jrmuizel:
What/where is `CGContextRef.h`?

IjzerenHein:
I looked at the `CGContextRef.h` file from the iOS 16 SDK from within XCode. Not sure where you can find it online. I'll create a gist for it:
https://gist.github.com/IjzerenHein/e99754c2a8e7e17450a4c77d86cd4c8d

PR #533, opened 2022-12-23
Add all color space names

AlexanderEkdahl:
Adds all of the non-deprecated color space names found at: https://developer.apple.com/documentation/coregraphics/cgcolorspace/color_space_names

This is required for EDR (Extended Dynamic Range) rendering.


jdm:
@bors-servo r+
Thanks!

PR #544, opened 2023-04-09
Update to use 2021 edition, and update deps

burtonageo:
I understand this is a big review, but I tried to change the minimal amount to get it compiling under the new circumstances.


jrmuizel:
Can you separate out the dependency updates?

jrmuizel:
Ideally each one separately

burtonageo:
I've moved the dependency upgrade to https://github.com/servo/core-foundation-rs/pull/545, so I can probably close this PR for now

mrobinson:
This is probably duplicated by #704 now.

PR #545, opened 2023-04-10
Update dependencies to latest versions

burtonageo:
This is a simplification of https://github.com/servo/core-foundation-rs/pull/544 to avoid the edition upgrade.


jrmuizel:
Can you split out each dependency update separately?

mrobinson:
The two new PRs are now landed, so this can be closed.

PR #548, opened 2023-04-25
Update `foreign-types` crate to v0.5

burtonageo:
Splits out the `foreign-types` part of https://github.com/servo/core-foundation-rs/pull/545


burtonageo:
Looks like it failed setting up the repository rather than when building

PR #549, opened 2023-04-25
Update `bitflags` to 2

burtonageo:
Splits out the `bitflags` part of https://github.com/servo/core-foundation-rs/pull/545

### DIFF
--- a/cocoa-foundation/Cargo.toml
+++ b/cocoa-foundation/Cargo.toml
@@ -14,7 +14,7 @@ default-target = "x86_64-apple-darwin"
 
 [dependencies]
 block = "0.1"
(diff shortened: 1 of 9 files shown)


mrobinson:
This causes semver checks to start failing because `fn from_bits_unchecked(...)` is no longer exposed on bitflags types.

mrobinson:
I have rebaesed this PR, so assuming it passes all CI jobs I'll land it. The only substantive change I have made to the original version is to specify `bitflags = "2"` for the version instead of the exact version. This should make it slightly more compatible with other projects that use different versions of bitflags.

mrobinson (approved):

PR #551, opened 2023-05-21
Add new functions in core-foundation-sys

michaelwright235:
I've added a lot of missing external functions, consts, etc. (only those that are available up to mac os 10.7) to core-foundation-sys crate. Almost all functions in the crate are now organized according to "Functions by task" in Apple docs.


michaelwright235:
I'll split this pull request in multiple ones, so it'd be easier to revise.

PR #552, opened 2023-05-23
sys: Add CFUserNotification

michaelwright235:
Adds `CFUserNotification` module into `core-foundation-sys` crate with all its constants and functions (https://developer.apple.com/documentation/corefoundation/cfusernotification?language=objc).
It also contains 3 plain inlined functions rewritten from CFUserNotification.h.
This API is available only on macOS, so an additional cfg flag was added in lib.rs.


michaelwright235:
@jdm Can you rerun the testing procedure? After the last commit it should work fine haha

PR #553, opened 2023-05-23
Bump macOS GitHub actions to macOS 11

michaelwright235:
macOS 10.15 runner image is removed since 8/30/22
(https://github.blog/changelog/2022-07-20-github-actions-the-macos-10-15-actions-runner-image-is-being-deprecated-and-will-be-removed-by-8-30-22/).
Using it makes GitHub Actions stuck forever.


jdm:
@bors-servo r+
Aha, that explains it. Thanks!

PR #554, opened 2023-05-24
Add binding for ScreenCaptureAccess

fcfangcc:
Add binding for [CGRequestScreenCaptureAccess](https://developer.apple.com/documentation/coregraphics/3656524-cgrequestscreencaptureaccess) and [CGPreflightScreenCaptureAccess](https://developer.apple.com/documentation/coregraphics/3656523-cgpreflightscreencaptureaccess) 

 to check if the app has screen recording permission on macOS

PR #560, opened 2023-05-29
Release for core-graphics  

fcfangcc:
Release core-graphics  version to 0.22.4

Include #554 , I urgently need to use it. Thanks 😃


jdm:
Published.

jdm:
I had to yank the- new version. This will need to be 0.23.0 instead due to the update to foreign-types that was present in this release.

jrmuizel:
Perhaps we should be running `cargo-semver-checks` in CI?

PR #562, opened 2023-05-29
Release cocoa patch to match up with latest core-graphics patch

### LINKED ISSUE #561: BUG | core-graphic patch broke cocoa due to lax semver
The recent patch to core-graphic is breaking cocoa@0.24.1 because cocoa relied on core-graphic 0.22, which resolved to the latest patch that was pushed ~3hrs ago.

I think cutting a patch for cocoa will be needed. Either to have the new code using the latest core-graphic, or restrict the core-graphic version to 0.22.1 is cool too. Otherwise, for project that has cocoa as transitive dependency pointing at 0.24, there's no straightforward way to override core-grahic to the previous version.

louisgv:
Related to https://github.com/servo/core-foundation-rs/pull/560 

resolves #561 

This release is needed for cocoa@0.24 to build


jdm:
We're going to need major version releases instead.

PR #575, opened 2023-06-01
sys: string: Add missing stuff

michaelwright235:
Adds `CFStringCompareFlags` constants, `CFStringEncoding` constants. Implements `CFStringNormalizationForm` type and its constants. Adds missing functions of `CFString` (they're sorted in Apple docs order) and implements `CFMutableStringRef` with its corresponding functions.


jrmuizel:
b742606a5c1e063ae5b2b23ecb05c017ea5a7bfc broke semver on core-foundation because UniChar is not exported by string.rs anymore.

michaelwright235:
@jrmuizel According to CFBase.h this type should belong to `base` module, so I've moved it there. It's a common type used by multiple modules. If it breaks semver I can revert this back.

PR #581, opened 2023-06-06
Fix compilation warnings

michaelwright235:
We get a couple of warnings during compilation, so I've fixed them.

### DIFF
--- a/core-graphics/src/event.rs
+++ b/core-graphics/src/event.rs
@@ -505,7 +505,7 @@ impl<'tap_life> CGEventTap<'tap_life> {
                     callback_ref: Box::from_raw(cbr),
                 })
             } else {
-                Box::from_raw(cbr);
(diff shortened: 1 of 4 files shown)


core-graphics/src/event.rs
```diff
@@ -505,7 +505,6 @@ impl<'tap_life> CGEventTap<'tap_life> {
                     callback_ref: Box::from_raw(cbr),
                 })
             } else {
-                Box::from_raw(cbr);
```
jdm, line 508: This is not useless; it prevents memory being leaked. We could assign it to _ if that makes the warning go away.

jdm (approved):

core-graphics/src/event.rs
michaelwright235: Yes, it'll make it go way. I'll change it.

PR #606, opened 2023-06-21
Bump core-foundation-sys to 0.9.0

michaelwright235:
A lot of stuff has been added since the last version, a minor version needs to be bumped.

### DIFF
--- a/core-foundation-sys/Cargo.toml
+++ b/core-foundation-sys/Cargo.toml
@@ -3,7 +3,7 @@ name = "core-foundation-sys"
 description = "Bindings to Core Foundation for macOS"
 homepage = "https://github.com/servo/core-foundation-rs"
 repository = "https://github.com/servo/core-foundation-rs"
-version = "0.8.4"
+version = "0.9.0"
 authors = ["The Servo Project Developers"]
 license = "MIT OR Apache-2.0"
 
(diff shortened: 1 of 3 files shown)


jrmuizel:
Are there actual semver breaks that requiring going to 0.9.0 vs 0.8.5? The semver CI suggests that 0.8.5 should be fine.

michaelwright235:
The only thing I concern of is this [3478886](https://github.com/servo/core-foundation-rs/pull/588/commits/34788869ec212947991ef6ba7938730c60a112a4)

michaelwright235:
@jrmuizel Should we bump the version to 0.8.5 then?

jrmuizel:
I wonder if the best thing to do here is to use conditional compilation to have both the right and wrong version of `CFComparatorFunction` and then when enough semver breaking stuff has accumulated we make the call and switch to 0.9 and remove the 0.8 code.

madsmtm:
I think that this kind of thing would constitute "acceptable breakage", since it is a bug fix for something that was clearly unsound before.

Do we have examples of dependencies that actually use the original function, that would be broken by this change?

jrmuizel:
> I think that this kind of thing would constitute "acceptable breakage", since it is a bug fix for something that was clearly unsound before, and someone getting a build error if they end up depending on this would be a good thing, since it could potentially be a security issue.

Yeah, I'm mostly convinced by this argument.

> Do we have examples of dependencies that actually use the original function, that would be broken by this change?

I did a quick check and didn't find anything.

Let's go with 0.8.5 and we can always pull it if someone complains about actual breakage.

michaelwright235:
I've created a separate #607 to bump the version to 0.8.5.

PR #608, opened 2023-06-30
feat: add link feature

crowlKats:
The reason for this is that weak linking required for Deno to bring back webgpu support, and as we use the wgpu crate which depends on core-foundation, this feature flag needs to be added to core-foundation as well, as we are not going to be building with nightly or unstable rust features.

### DIFF
--- a/cocoa/Cargo.toml
+++ b/cocoa/Cargo.toml
@@ -15,8 +15,13 @@ default-target = "x86_64-apple-darwin"
 block = "0.1"
 bitflags = "1.0"
 libc = "0.2"
-cocoa-foundation = { path = "../cocoa-foundation", version = "0.1" }
-core-foundation = { path = "../core-foundation", version = "0.9" }
-core-graphics = { path = "../core-graphics", version = "0.23" }
+cocoa-foundation = { default-features = false, path = "../cocoa-foundation", version = "0.1" }
+core-foundation = { default-features = false, path = "../core-foundation", version = "0.9" }
+core-graphics = { default-features = false, path = "../core-graphics", version = "0.23" }
 foreign-types = "0.5"
 objc = "0.2.3"
+
+[features]
+default = ["link"]
+# Disable to manually link. Enabled by default.
+link = ["core-foundation/link", "cocoa-foundation/link", "core-graphics/link"]
--- a/cocoa-foundation/Cargo.toml
+++ b/cocoa-foundation/Cargo.toml
@@ -15,6 +15,12 @@ default-target = "x86_64-apple-darwin"
 block = "0.1"
 bitflags = "1.0"
 libc = "0.2"
-core-foundation = { path = "../core-foundation", version = "0.9" }
-core-graphics-types = { path = "../core-graphics-types", version = "0.1" }
+core-foundation = { default-features = false, path = "../core-foundation", version = "0.9" }
+core-graphics-types = { default-features = false, path = "../core-graphics-types", version = "0.1" }
 objc = "0.2.3"
+
+[features]
+default = ["link"]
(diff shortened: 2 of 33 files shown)


jrmuizel:
Could you elaborate a bit more the context for this change?

crowlKats:
As I have stated, this is necessary for Deno. This is because the WebGPU implementation was removed due to static linking causing massive startup time issues, and as the usage of said web API is not quite common yet, we decided to temporarily remove the implementation. Now we are looking into bringing WebGPU back into Deno, and to do that we need to weak link frameworks. We have started doing this already in the gfx-rs repos (https://github.com/gfx-rs/metal-rs/pull/271, https://github.com/gfx-rs/wgpu/pull/3853, https://github.com/gfx-rs/wgpu/pull/3897), and the main & remaining frameworks used are coming from core-foundation.

jrmuizel:
I still don't understand. Perhaps this is me not understanding what exactly`#[link]` does. How are the symbols resolved without it?

Also, macOS doesn't support statically linking against the system libraries so I don't understand what you mean when you say static linking caused a massive startup time issues.

littledivy:
Hey @jrmuizel 

With #[link], macOS will always load the framework at startup (dyld). We want to use [`-weak_framework`](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPFrameworks/Concepts/WeakLinking.html) to avoid this behaviour and save some startup time.

cocoa/Cargo.toml
```diff
@@ -20,3 +20,8 @@ core-foundation = { path = "../core-foundation", version = "0.9" }
 core-graphics = { path = "../core-graphics", version = "0.23" }
 foreign-types = "0.5"
 objc = "0.2.3"
+
+[features]
+default = ["link"]
+# Use static linking for frameworks. Disale to manually link. Enabled by default.
```
littledivy, line 26:
```suggestion
# Disable to manually link. Enabled by default.
```
Remove 'static linking'. Not the right word

madsmtm:
So ideally, I think we should work towards [having the option in the language](https://internals.rust-lang.org/t/support-for-macos-weak-linking/13796) (`#[link(name = "Foundation", kind = "weak_framework")]`), such that the linking could still be specified by the library.

But since that's a bit farther away, I think this is a fine stopgap solution.

crowlKats:
apologies for the delay. @jrmuizel I have squashed all the commits.

crowlKats:
@jrmuizel when can we expect new releases for the crates?

jrmuizel:
Which crates do you all need releases for?

crowlKats:
For now `core-graphics-types` would suffice, and we can wait for the other ones

crowlKats:
I see; would it be possible to then publish those as well, or is there some blocker?

crowlKats:
@jrmuizel I looked through the repo and that spot seems the only one (as all other packages depend on non-fully qualified semvers). I am not really sure I understand how the release system works for this repo though, so I might be missing something.

jrmuizel:
@crowlKats I made a new release of core-graphics-types

madsmtm:
I'll note that `weak_framework` is [recently](https://github.com/rust-lang/rust/pull/118448) possible on nightly:
```rust
#![feature(link_arg_attribute)]

#[link(name = "-weak_framework", kind = "link-arg", modifiers = "+verbatim")]
#[link(name = "CoreFoundation", kind = "link-arg", modifiers = "+verbatim")]
extern "C" {}
```

This works even if declared in a binary crate that depends on something that has the normal `#[link(name = "CoreFoundation", kind = "framework")]`.

Given that, maybe we should consider removing this feature again?

PR #609, opened 2023-07-03
core-foundation-sys: Enable `no_std` environment

michaelwright235:
Since the purpose of this crate is the bindings to an external library and they're pretty much complete, we can see that the only thing we use in libstd are the pointer types and `std::cmp`. In fact they're defined in libcore and then exported through libstd. I doubt that we will ever need any libstd only stuff such as heap allocation, vectors, hashmaps, etc. I propose making core-foundation-sys a `no_std` crate. It maybe useful in some rare situations. For instance, windows-sys crate (Microsoft's official Rust bindings) is no_std and it can be used for writing kernel drivers.

### DIFF
--- a/core-foundation-sys/src/array.rs
+++ b/core-foundation-sys/src/array.rs
@@ -7,7 +7,7 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
-use std::os::raw::c_void;
+use core::ffi::c_void;
 
 use base::{CFRange, CFIndex, CFAllocatorRef, CFTypeID, Boolean, CFComparatorFunction};
 use string::CFStringRef;
--- a/core-foundation-sys/src/attributed_string.rs
+++ b/core-foundation-sys/src/attributed_string.rs
@@ -7,7 +7,7 @@
 // option. This file may not be copied, modified, or distributed
 // except according to those terms.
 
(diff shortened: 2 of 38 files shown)


madsmtm (approved):
:+1:

Beware that this bumps MSRV to 1.64.

michaelwright235:
Maybe it's acceptable because `core-foundation` that depends on this crate already has minimum Rust version of 1.64.0 ([using of std::ffi::Cstr](https://github.com/servo/core-foundation-rs/blob/8f7400bd4361a5184f1ef2b2fdabe436f526f03d/core-foundation/src/string.rs#L22C1-L22C20)).

michaelwright235:
Hmm, it looks like this import has been added a couple of years ago, but rust docs says it was added in 1.64.0

mrobinson:
This was done in #692 in the end.

waywardmonkeys:
If we treat this like an issue rather than a PR, there's still 3 things outstanding:

* `std:::cmp::Ordering` is used in `core-foundation-sys/src/base.rs`
* `std::ptr::null()` is used in `core-foundation-sys/src/bundle.rs`
* `no_std` isn't enabled specifically.

If @michaelwright235 would like to rebase this to do that, that's great ... or we can open a new PR that does so?

PR #612, opened 2023-07-07
Drop #[no_mangle] from cg_event_tap_callback_internal

jrmuizel:
This allows multiple versions of core-graphics to be linked into the same binaries without getting "duplicate symbol: cg_event_tap_callback_internal"


jrmuizel:
Previously the only major version that included this symbol was 0.22 now that 0.23 also has it we're much more likely to get in trouble.

PR #616, opened 2023-07-11
Enable merge queue

mukilan:


.github/workflows/rust.yml
```diff
@@ -39,7 +41,7 @@ jobs:
     - name: Run tests
       run: cargo test --verbose
   build_result:
-    name: homu build finished
+    name: Result
     runs-on: ubuntu-latest
     needs:
```
mukilan, line 46: Do we need to include the other jobs(msrv & semver) here in 'needs' clause?

mrobinson: Yeah, it looks like we might need that.

mrobinson (approved):

PR #621, opened 2023-07-25
Allow to build against bitflags 2.x

glandium:


glandium:
Mmm as this changes the traits derived for the public types, this doesn't quite work.

PR #623, opened 2023-07-29
ci: Remove Travis configs.

waywardmonkeys:
These haven't been used in a year. Also, change the build status badges to point to GitHub Actions.


jrmuizel:
I'd actually like to get travis running again because it supports older versions of macOS

PR #628, opened 2023-07-31
Replace objc with objc

madsmtm:
Redo of https://github.com/servo/core-foundation-rs/pull/513, GitHub was giving me trouble.

### DIFF
--- a/core-graphics-types/src/lib.rs
+++ b/core-graphics-types/src/lib.rs
@@ -8,7 +8,7 @@
 // except according to those terms.
 
 extern crate core_foundation;
-extern crate libc;
+extern crate objc_encode;
 
 pub mod base;
 pub mod geometry;
--- a/core-graphics-types/src/lib.rs
+++ b/core-graphics-types/src/lib.rs
@@ -8,7 +8,7 @@
 // except according to those terms.
 
 extern crate core_foundation;
-extern crate objc2_encode;
+extern crate objc2;
 
 pub mod base;
 pub mod geometry;
--- a/core-graphics-types/src/lib.rs
+++ b/core-graphics-types/src/lib.rs
@@ -8,7 +8,7 @@
 // except according to those terms.
 
 extern crate core_foundation;
-extern crate objc_encode;
+extern crate objc2_encode;
 
 pub mod base;
 pub mod geometry;
--- a/core-graphics-types/src/lib.rs
+++ b/core-graphics-types/src/lib.rs
@@ -8,6 +8,7 @@
 // except according to those terms.
 
 extern crate core_foundation;
+#[cfg(feature = "objc2")]
 extern crate objc2;
 
 pub mod base;
--- a/core-text/src/font.rs
+++ b/core-text/src/font.rs
@@ -206,7 +206,7 @@ pub fn new_ui_font_for_language(
     unsafe {
         let font_ref = CTFontCreateUIFontForLanguage(
             ui_type,
-            size,
+            size as CGFloat,
             language
                 .as_ref()
                 .map(|x| x.as_concrete_TypeRef())
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -34,7 +34,7 @@ jobs:
     - name: Install toolchain
       uses: actions-rs/toolchain@v1
       with:
-        toolchain: 1.56.1
+        toolchain: '1.60'
         override: true
     - name: Build
       run: cargo build --verbose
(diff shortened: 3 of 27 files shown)


madsmtm:
(That the semver checks are failing is expected, this _is_ a breaking change).

nicoburns:
@madsmtm Do you have a read on how this crate differs from the relevant higher-level crates in the objc2 family? (are there equivalent crate(s)). Would it make sense to look at more heavily adopting some of those crates? Or even outright replacing (some of?) the crates in this repo with ones from https://github.com/madsmtm/objc2 ?

madsmtm:
> Do you have a read on how this crate differs from the relevant higher-level crates in the objc2 family? (are there equivalent crate(s)). Would it make sense to look at more heavily adopting some of those crates? Or even outright replacing (some of?) the crates in this repo with ones from https://github.com/madsmtm/objc2 ?

There's some fairly old discussion of a few lower-level details in https://github.com/servo/core-foundation-rs/pull/513, but yeah, nowadays we have `objc2-foundation`, `objc2-app-kit` and `objc2-quartz-core`, which completely replace the functionality in `cocoa`, along with providing:
- Actual type safety (everything isn't just an `id` pointer)
- Proper memory management (no longer need to do `retain`/`release` calls, autorelease pools aren't as important)
- Automatically generated bindings to almost everything in those frameworks

nicoburns:
A couple of followup questions:

- Does this extra type safety and generated bindings have compile time and/or code size implications?

- Are there bindings for CoreText somewhere in the objc2 family of crates? Those seem to be a heavily used part of this repo, but I couldn't find them for objc2. And if not, are there other frameworks for which bindings are currently missing?

madsmtm:
> Does this extra type safety and generated bindings have compile time and/or code size implications?

They do have a compile-time cost, but it's quite low by now if you only enable the features you need (which is the default behaviour unless you use the `"all"` feature).

Wrt. code-size, as long as you use LTO, there shouldn't be a difference. I'm considering adding `#[inline]` to ensure that this happens without, but I'm slightly unsure of the compile-time cost of doing so, so haven't done it yet.

> Are there bindings for CoreText somewhere in the objc2 family of crates? Those seem to be a heavily used part of this repo, but I couldn't find them for objc2. And if not, are there other frameworks for which bindings are currently missing?

CoreText is in the category of a `CoreFoundation`-like framework, these aren't supported yet because they don't actually use Objective-C, and the memory management scheme is completely different, see https://github.com/madsmtm/objc2/issues/556.

core-text/src/font.rs
```diff
@@ -206,7 +206,7 @@ pub fn new_ui_font_for_language(
     unsafe {
         let font_ref = CTFontCreateUIFontForLanguage(
             ui_type,
-            size,
+            size as CGFloat,
```
waywardmonkeys, line 209: This is needed if you build for a 32 bit target, but are there any left (that support CoreText)?

core-graphics-types/src/lib.rs
```diff
@@ -8,7 +8,8 @@
 // except according to those terms.
 
 extern crate core_foundation;
-extern crate libc;
+#[cfg(feature = "objc2")]
```
waywardmonkeys, line 11: I think there's value in having a standalone PR that adds the `Encode` traits for `objc2` to `core-graphics-types` via a feature. Would that help for some of what you maintain in the `objc2` family of crates?

It is separate from converting the `cocoa` and `cocoa-foundation` crates, right?

waywardmonkeys (commented):
I think it is worth extracting at least 2 parts of this out (the changes to `core-text` and `core-graphics-types`) and considering them separately from the changes to `cocoa` and `cocoa-foundation`.

madsmtm:
> I think it is worth extracting at least 2 parts of this out (the changes to `core-text` and `core-graphics-types`) and considering them separately from the changes to `cocoa` and `cocoa-foundation`.

Gotta admit, I'm not really motivated to work on this repo rn (though I may become so in the future, esp. considering you just joined as maintainer).

In any case, I've extracted the first commit into https://github.com/servo/core-foundation-rs/pull/702, as I indeed think that change is valuable on its own, and does make things easier to review here.

core-graphics-types/src/lib.rs
madsmtm: Hmm, I _think_ it would help, though I'll have to mull it over; it kinda depends on how long it will take me to get to doing https://github.com/madsmtm/objc2/issues/556 (in which case all the types in this repo will just be autogenerated), and I suspect a lot of people would be annoyed at the extra dependency.

core-text/src/font.rs
madsmtm: Rust does still support some 32bit targets like `i686-apple-darwin`, `armv7s-apple-ios` and `i386-apple-ios`. And those do support CoreText.

wusyong:
@madsmtm I think it's okay to wait. Perhaps Servo should use `objc2`'s framework bindings, just as Servo recently migrated from `winapi` to `windows-sys`.
Though several frameworks Servo uses don't exist yet, we could discuss this in future issues.

PR #631, opened 2023-08-08
Update `bitflags` to 2.2.1

daxpedda:
Continuation of #549.


jrmuizel:
I'm not sure that it's worth breaking semver for this. Maybe it should wait until there's other reasons to break semver?

daxpedda:
I agree, this is probably not worth breaking over.

Issue #648, opened 2023-12-04
Safe CFRunLoop methods accept and dereference a raw pointer


kevinmehall:
The CFRunLoop methods `add_timer`, `remove_timer` `contains_source`, `add_source`, `remove_source`, `contains_observer`, `add_observer`, `remove_observer` accept an argument `mode` of type `CFRunLoopMode`, which is a type alias for `*const __CFString`. This parameter is passed on to the underlying functions which dereference it, so these methods are unsound because nothing prevents safe code from passing an invalid pointer.

I confirmed that the following line without `unsafe` segfaults:
```rust
run_loop.add_source(&source.unwrap(), 0x1234 as *mut _)
```

Ironically, correct usage requires `unsafe`, because the `kCFRunLoopDefaultMode` and `kCFRunLoopCommonModes` that are normally passed here are `extern static`s that requires `unsafe` to access.

PR #650, opened 2023-12-04
Make CFRunLoopMode a safe wrapper with a lifetime

kevinmehall:
Fixes https://github.com/servo/core-foundation-rs/issues/648: the safe methods that take `CFRunLoopMode` are currently unsound because it is a raw pointer that is not necessarily valid.

As a bonus, this allows use of `kCFRunLoopCommonModes` and `kCFRunLoopDefaultMode` constants without `unsafe`.

I considered just making `mode` a `&CFString`, but it seems that the [implementation compares the pointer values](https://github.com/apple-oss-distributions/CF/blob/dc54c6bb1c1e5e0b9486c1d26dd5bef110b20bf3/CFRunLoop.c#L2824), so this way ensures that the constants are passed as the pointers it expects. I don't understand the purpose or usage of `mode` very well, so correct me if this is not ideal.

### DIFF
--- a/core-foundation/src/filedescriptor.rs
+++ b/core-foundation/src/filedescriptor.rs
@@ -154,7 +154,7 @@ mod test {
         let run_loop = CFRunLoop::get_current();
         let source = CFRunLoopSource::from_file_descriptor(&cf_fd, 0);
         assert!(source.is_some());
-        run_loop.add_source(&source.unwrap(), CFRunLoopMode::default());
+        run_loop.add_source(&source.unwrap(), CFRunLoopMode::default_mode());
 
         info.value = 0;
         cf_fd.enable_callbacks(kCFFileDescriptorReadCallBack);
--- a/core-foundation/src/runloop.rs
+++ b/core-foundation/src/runloop.rs
@@ -29,12 +29,12 @@ impl<'a> CFRunLoopMode<'a> {
     }
 
     #[doc(alias = "kCFRunLoopCommonModes")]
-    pub fn common() -> CFRunLoopMode<'static> {
+    pub fn common_modes() -> CFRunLoopMode<'static> {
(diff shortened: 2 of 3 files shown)


waywardmonkeys:
Hello! Sorry for the long delay.

Would you be able to rebase this forward and fix the merge conflicts? If so, we'd like to review it and see about landing it.

kevinmehall:
Rebased

kevinmehall:
Clippy doesn't like methods called `default`:

```
error: method `default` can be confused for the standard trait method `std::default::Default::default`
  --> core-foundation/src/runloop.rs:37:5
   |
37 | /     pub fn default() -> CFRunLoopMode<'static> {
38 | |         unsafe { CFRunLoopMode(kCFRunLoopDefaultMode, PhantomData) }
39 | |     }
   | |_____^
   |
   = help: consider implementing the trait `std::default::Default` or choosing a less ambiguous method name
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#should_implement_trait
   = note: `-D clippy::should-implement-trait` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(clippy::should_implement_trait)]`
```

so I renamed the constructors to `CFRunLoopMode::default_mode()` and `CFRunLoopMode::common_modes()`. But if you think that's too redundant, we could also ignore the clippy lint.

waywardmonkeys:
I think the naming, while redundant, is fine.

kevinmehall:
Pushed again to fix `rustfmt` on a line that `default_mode` pushed over the length limit.

(FYI, there's a GitHub Actions setting for "Require approval for first-time contributors who are new to GitHub" that is less annoying than the default "Require approval for first-time contributors ")

PR #652, opened 2024-01-02
Fix  TCFType macros for generic types

tmandry:
`impl_TCFType!` accepts generic types but some of its impls are missing the generics. They work on types with defaulted generics, but only in the default case. This PR fixes that.

It also updates `declare_TCFType!` to accept generics as a convenience. We use one PhantomData per parameter, as is already done in `impl_TCFType!`.

In addition, some warnings are silenced, and the macros no longer depend on any traits being in scope in the caller.

### DIFF
--- a/core-foundation/src/lib.rs
+++ b/core-foundation/src/lib.rs
@@ -53,15 +53,24 @@ macro_rules! declare_TCFType {
         $(#[$doc:meta])*
         $ty:ident, $raw:ident
     ) => {
-        $(#[$doc])*
-        pub struct $ty($raw);
+        declare_TCFType!($(#[$doc])* $ty<>, $raw);
+    };
 
-        impl Drop for $ty {
+    (
+        $(#[$doc:meta])*
+        $ty:ident<$($p:ident $(: $bound:path)*),*>, $raw:ident
+    ) => {
+        $(#[$doc])*
+        pub struct $ty<$($p $(: $bound)*),*>($raw, $(::std::marker::PhantomData<$p>),*);
+
+        #[allow(unused_imports)]
+        impl<$($p $(: $bound)*),*> Drop for $ty<$($p),*> {
             fn drop(&mut self) {
+                use $crate::base::TCFType;
                 unsafe { $crate::base::CFRelease(self.as_CFTypeRef()) }
             }
         }
-    }
+    };
 }
 
 /// Provide an implementation of the [`TCFType`] trait for the Rust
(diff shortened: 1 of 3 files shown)


tmandry:
This is needed to create a type-safe generic abstraction over `AXValue`; see value.rs in https://github.com/eiz/accessibility/pull/5.

tmandry:
Friendly ping @jdm, would you or another maintainer be willing to take a look at this?

waywardmonkeys:
@tmandry I'd like to get this unblocked for you. Could you rebase to current `main`?

core-foundation/src/lib.rs
```diff
@@ -121,39 +132,48 @@ macro_rules! impl_TCFType {
...
         }
 
-        impl Eq for $ty { }
+        impl<$($p $(: $bound)*),*> Eq for $ty<$($p),*> { }
 
-        unsafe impl<'a> $crate::base::ToVoid<$ty> for &'a $ty {
+        #[allow(unused_imports)]
+        unsafe impl<'a, $($p $(: $bound)*),*> $crate::base::ToVoid<$ty<$($p),*>> for &'a $ty<$($p),*> {
             fn to_void(&self) -> *const ::std::os::raw::c_void {
+                use $crate::base::TCFType;
```
waywardmonkeys, line 160: Please merge with the line below.

core-foundation/src/lib.rs
```diff
@@ -121,39 +132,48 @@ macro_rules! impl_TCFType {
...
                 use $crate::base::TCFTypeRef;
                 self.as_concrete_TypeRef().as_void_ptr()
             }
         }
 
-        unsafe impl $crate::base::ToVoid<$ty> for $ty {
+        #[allow(unused_imports)]
+        unsafe impl<$($p $(: $bound)*),*> $crate::base::ToVoid<$ty<$($p),*>> for $ty<$($p),*> {
             fn to_void(&self) -> *const ::std::os::raw::c_void {
+                use $crate::base::TCFType;
```
waywardmonkeys, line 169: Merge with the line below here as well.

waywardmonkeys (commented):
Would it be possible to add a test so that it at least remains in a compiling state?

tmandry:
Thanks for taking a look @waywardmonkeys. Should be ready now.

waywardmonkeys (approved):
Thanks! I think this looks great! Will leave open for a day or so in case others want to review.

PR #653, opened 2024-01-13
fix: Correcting errors in window CGWindowListOption values

nashaofu:
https://developer.apple.com/documentation/coregraphics/quartz_window_services/window_image_types?language=objc
image


madsmtm (approved):
Note that these are also defined (correctly) over in

https://github.com/servo/core-foundation-rs/blob/d4ce710182f1756c9d874ab917283fe1a1b7a011/core-graphics/src/display.rs#L33-L38

nashaofu:
> Note that these are also defined (correctly) over in
> 
> https://github.com/servo/core-foundation-rs/blob/d4ce710182f1756c9d874ab917283fe1a1b7a011/core-graphics/src/display.rs#L33-L38

Why are there so many duplicate definitions in the project? Is there any problem if I remove it?

jdm (approved):
Thanks!

jdm:
Cleaning up duplicate definitions would be great!

PR #657, opened 2024-03-13
feat(CGDisplay): add displays_with_point and displays_with_rect

mattjbray:
Add [`CGDisplay::displays_with_point`](https://developer.apple.com/documentation/coregraphics/1454385-cggetdisplayswithpoint) and [`CGDisplay::displays_with_rect`](https://developer.apple.com/documentation/coregraphics/1456071-cggetdisplayswithrect).

Uses the pattern from `CGDisplay::active_displays` to first get the count of matching displays, then intialize the result Vec with `vec![0; count as usize]`.

### DIFF
--- a/core-graphics/src/display.rs
+++ b/core-graphics/src/display.rs
@@ -929,3 +929,71 @@ extern "C" {
         imageOptions: CGWindowImageOption,
     ) -> crate::sys::CGImageRef;
 }
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn test_display_count_with_point() {
+        let result = CGDisplay::display_count_with_point(CGPoint::new(0., 0.));
+        assert!(result.is_ok());
+    }
+
+    #[test]
+    fn test_displays_with_point_0() {
+        let result = CGDisplay::displays_with_point(CGPoint::new(0., 0.), 0);
+        assert!(result.is_ok());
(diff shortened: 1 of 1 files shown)


jrmuizel:
Can you add some tests?

mattjbray:
> Can you add some tests?

Added some simple tests. I'm not sure how to test anything more specific without mocking the displays somehow - any pointers on how to do that would be appreciated.

mattjbray:
I've also added `#[inline]` attributes to match all the other fns in this module.

core-graphics/src/display.rs
```diff
@@ -141,6 +141,90 @@ impl CGDisplay {
...
 
+    /// Return the number of online displays with bounds that include the
+    /// specified point.
+    #[inline]
+    pub fn display_count_with_point(point: CGPoint) -> Result<u32, CGError> {
+        let mut matching_display_count: u32 = 0;
+        let result = unsafe {
+            CGGetDisplaysWithPoint(point, 0, ptr::null_mut(), &mut matching_display_count)
+        };
+        if result == 0 {
```
waywardmonkeys, line 152: Would you mind comparing against `kCGErrorSuccess` in these cases?

mattjbray: Done (and rebased on latest `main`).

waywardmonkeys (approved):

PR #658, opened 2024-03-20
Rename `master` branch to `main`

mrobinson:
This is in line with a decision made about the Servo organization by the Servo TSC. We consider `main` to be shorter, easier to type, and more modern, so we are standardizing on this branch name.


mrobinson (commented):
Immediately before landing this, the branch name will need to be updated in the GitHub UI.

mukilan (approved):

PR #660, opened 2024-04-03
NSColor: add AppKit-defined colors

mattjbray:
Add AppKit-defined UI Element Colors and Standard Colors.

See https://developer.apple.com/documentation/appkit/nscolor#1651631

### DIFF
--- a/cocoa/src/appkit.rs
+++ b/cocoa/src/appkit.rs
@@ -5364,4 +5364,108 @@ mod test {
             let _nsApp = NSApp();
         }
     }
+
+    #[test]
+    pub fn test_nscolor_appkit_colors() {
+        use crate::base::nil;
+        unsafe {
+            // Check all colors are defined
+
+            // Label Colors
+            let _ = NSColor::labelColor(nil);
+            let _ = NSColor::secondaryLabelColor(nil);
+            let _ = NSColor::tertiaryLabelColor(nil);
+            let _ = NSColor::quaternaryLabelColor(nil);
+
+            // Text Colors
+            let _ = NSColor::textColor(nil);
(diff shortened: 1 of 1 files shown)


mattjbray:
Looks like `quaternarySystemFillColor` et al are only available on macOS 14.0+, but the tests run on `os: [macos-11.0, macos-12, macos-13]`.

Is there any preprocessor/directive already in use to detect what version of macOS is being built against? Or should I just remove them?

waywardmonkeys:
Part of the issue here is that you're running tests that need the new symbols, but if you didn't do that, then things would compile, leaving up to the person using the crate to sort it out. I'm not sure what's better.

mattjbray:
An option is to add a `macos_14` cargo feature (set by default) and disable/enable it in the test matrix. I can have a go at that if you want.

mattjbray:
There is some precedent for using cargo features:
* https://github.com/servo/core-foundation-rs/blob/b2fdaf4132a8fff7aadd6885ec7c570296a664ea/core-foundation-sys/Cargo.toml#L21-L25
* https://github.com/servo/core-foundation-rs/blob/b2fdaf4132a8fff7aadd6885ec7c570296a664ea/core-foundation/Cargo.toml#L28-L32
* https://github.com/servo/core-foundation-rs/blob/b2fdaf4132a8fff7aadd6885ec7c570296a664ea/core-graphics/Cargo.toml#L27-L30
* https://github.com/servo/core-foundation-rs/blob/b2fdaf4132a8fff7aadd6885ec7c570296a664ea/core-text/Cargo.toml#L25-L29

I was thinking along the lines of how it's done in `core-text`: include the feature by default, and you can exclude it if building on older osx versions.

PR #662, opened 2024-04-04
Bump the patch version.

jrmuizel:
The change to kCGWindowListOptionAll was technically semver breaking but the constants were wrong before so it should be fine.


mstange (approved):
Seems a bit risky to do this as a point release but it's your call.

PR #663, opened 2024-04-04
Update/remove obsolete badges and update link

atouchet:


mrobinson (approved):
This looks good to me. If we ever re-enable Travis support we can add these back.

PR #665, opened 2024-04-05
Fixes #661

adamnemecek:


mattjbray:
See also https://github.com/servo/core-foundation-rs/pull/492

adamnemecek:
I guess that also works. It is disheartening that that PR been open since 2022.

adamnemecek:
The problem is that currently the api does not allow for consuming events.

PR #667, opened 2024-04-08
Upgrade to bitflags 2

djc:
A rebased version of #549 (retaining authorship from the OP).

cc @mrobinson following up on your plan from https://github.com/servo/core-foundation-rs/pull/521#issuecomment-2042258249.

### DIFF
--- a/core-text/src/font.rs
+++ b/core-text/src/font.rs
@@ -944,7 +944,8 @@ fn out_of_range_variations() {
     let small = new_ui_font_for_language(kCTFontSystemDetailFontType, 19., None);
 
     let axes = small.get_variation_axes();
-    if macos_version() < (10, 12, 0) {
+    let version = dbg!(macos_version());
+    if version < (10, 12, 0) {
         assert!(axes.is_none());
         return;
     }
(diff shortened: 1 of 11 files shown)


core-text/src/font.rs
```diff
@@ -1006,9 +1009,41 @@ fn out_of_range_variations() {
...
+            let tag = axis
+                .find(unsafe { kCTFontVariationAxisIdentifierKey })
+                .unwrap();
+            let max = axis
+                .find(unsafe { kCTFontVariationAxisMaximumValueKey })
+                .unwrap()
+                .downcast::<CFNumber>()
+                .unwrap()
+                .to_f64()
+                .unwrap();
```
mrobinson, line 1030: It looks like this code is unrelated?

mrobinson (commented):
Thanks for the rebase. It looks like this version contains some unrelated code, so I have just rebased the original PR and will merge there.

djc:
It contains code to make the tests pass on macOS 14 (which includes my dev machine). That still seems useful?

mrobinson:
> It contains code to make the tests pass on macOS 14 (which includes my dev machine). That still seems useful?

Yes, it could definitely be useful. Is it related to the bitflags changes though? Would you mind opening a new PR with these changes separating it out from the bitflags work?

PR #668, opened 2024-04-09
Fix support for macOS 14

djc:
@mrobinson as requested in https://github.com/servo/core-foundation-rs/pull/667#issuecomment-2044482894.

### DIFF
--- a/core-text/src/font.rs
+++ b/core-text/src/font.rs
@@ -982,11 +982,13 @@ fn out_of_range_variations() {
     let var_desc = variation_font.copy_descriptor();
     let var_attrs = var_desc.attributes();
     dbg!(&var_attrs);
-    // attributes greater than max are dropped on macOS <= 11
-    // on macOS 12 they seem to be preserved as is.
+
+    // Handling of attributes greater than max changed between versions
     let var_attrs = var_attrs.find(variation_attribute);
-    if version >= (12, 0, 0) && version < (13, 0, 0) {
-        check_attrs(var_attrs, axes);
+    if version >= (14, 0, 0) {
+        check_attrs(0., var_attrs, axes);
+    } else if version >= (12, 0, 0) && version < (13, 0, 0) {
+        check_attrs(1., var_attrs, axes);
     } else if version >= (10, 15, 0) {
         assert!(var_attrs.is_none());
     } else {
@@ -995,6 +997,7 @@ fn out_of_range_variations() {
     }
 
     fn check_attrs(
+        clamp_diff: f64,
         var_attrs: Option<ItemRef<'_, CFType>>,
         axes: CFArray<CFDictionary<CFString, CFType>>,
     ) {
@@ -1021,7 +1024,7 @@ fn out_of_range_variations() {
                 .to_f64()
                 .unwrap();
 
-            let expected = max + 1.;
+            let expected = max + clamp_diff;
             assert_eq!(
                 val, expected,
                 "axis {:?} = {:?} (expected {:?})",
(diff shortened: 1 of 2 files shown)


djc:
@mrobinson friendly ping? Do you have a plan for the release?

core-text/src/font.rs
```diff
@@ -983,7 +984,9 @@ fn out_of_range_variations() {
     // attributes greater than max are dropped on macOS <= 11
     // on macOS 12 they seem to be preserved as is.
     let var_attrs = var_attrs.find(variation_attribute);
-    if macos_version() >= (12, 0, 0) && macos_version() < (13, 0, 0) {
+
+    // Just guessing -- this passed on (14, 4, 1)
+    if version >= (14, 0, 0) {
```
mrobinson, line 989: Changing this first case makes the comment above this not make any sense.

core-text/src/font.rs
```diff
@@ -944,7 +944,8 @@ fn out_of_range_variations() {
     let small = new_ui_font_for_language(kCTFontSystemDetailFontType, 19., None);
 
     let axes = small.get_variation_axes();
-    if macos_version() < (10, 12, 0) {
+    let version = dbg!(macos_version());
```
mrobinson, line 947: I think you can remove `dbg!(...)` here. It doesn't add too much when you aren't working on this particular section of the code.

core-text/src/font.rs
```diff
@@ -1006,9 +1009,41 @@ fn out_of_range_variations() {
...
+                .downcast::<CFNumber>()
+                .unwrap()
+                .to_f64()
+                .unwrap();
+
+            let expected = max + 1.;
+            assert_eq!(
+                val, expected,
+                "axis: {:?} = {:?} (not {:?})",
+                tag, val, expected
```
mrobinson, line 1043: It's pretty hard to tell from this change, but is the only difference the lack of the assertion? If that's the case, I just add the conditional around the assertion and add a comment explaining why it doesn't apply.

mrobinson (changes requested):
Sorry for the delay in reviewing. Just a few questions about this...

core-text/src/font.rs
djc: There's a bunch of instances of `dbg!()` already in this test, which was really useful when this test "suddenly" started failing due to me being on a newer OS version, so I'd like to keep this around (which helps see on test failure what case is being hit).

(In general, I think it's pretty common to leave `println!()` or `dbg!()` calls in test code, since you'll usually only see it on test failure anyway.)

core-text/src/font.rs
djc: Fair -- I updated the comment to be more general.

core-text/src/font.rs
djc: I split my changes so they're hopefully easier to review on a commit-by-commit basis, and extracted a function with a parameter instead of duplicating this.

Hope that helps.

core-text/src/font.rs
```diff
@@ -980,10 +982,25 @@ fn out_of_range_variations() {
...
+    if version >= (14, 0, 0) {
+        check_attrs(0., var_attrs, axes);
+    } else if version >= (12, 0, 0) && version < (13, 0, 0) {
+        check_attrs(1., var_attrs, axes);
+    } else if version >= (10, 15, 0) {
+        assert!(var_attrs.is_none());
+    } else {
+        let var_attrs = var_attrs.unwrap().downcast::<CFDictionary>().unwrap();
+        assert!(var_attrs.is_empty());
+    }
```
mrobinson, line 997: Thanks, this is a nice improvement!

mrobinson (approved):

djc:
Do you have a plan for release?

mrobinson:
> Do you have a plan for release?

This just fixes a test, correct? Do we need to make a new release considering that `core-foundation-rs` still compiles and functions properly on MacOS 14?

djc:
I meant for #666 and #667.

PR #670, opened 2024-04-16
fix(core-graphics): add composite events to CGEventType

pewsheen:
I need to listen to a `NX_SYSDEFINED` event. I found it was not coded in the `CGEventType` enum but on the http://opensource.apple.com/source/IOHIDFamily/IOHIDFamily-700/IOHIDSystem/IOKit/hidsystem/IOLLEvent.h

### DIFF
--- a/core-graphics/src/event.rs
+++ b/core-graphics/src/event.rs
@@ -126,6 +126,11 @@ pub enum CGEventType {
     KeyUp = 11,
     FlagsChanged = 12,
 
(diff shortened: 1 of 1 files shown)


wusyong:
It seems Tauri applied some workaround to resolve this. Perhaps this one can be closed.

pewsheen:
Eventually, we linked the framework by ourselves, so this is no longer required.

PR #489, opened 2021-12-23
Update NSAppKitVersionNumber

youngsing:
Version Number get from macOS 12.1 and Xcode 13.2 

Screen Shot 2021-12-23 at 15 51 50

PR #673, opened 2024-04-27
cocoa: update NSAppKitVersionNumbers

santikid:
Add NSAppKitVersionNumbers according to NSApplication.h.
#489

NSApplication.h screenshot


nicoburns (approved):
This is now slightly out of date (latest version at time of writing appears to be 14.5) but seems like a strict improvement on what we have.

https://developer.apple.com/documentation/appkit/nsappkitversionnumber?language=objc

PR #678, opened 2024-05-19
fix: return preflightscreencapture as true for all non zero values

### LINKED ISSUE #677: ScreenCaptureAccess.preflight returns incorrect value for Intel based 64bit Mac
The function ScreenCaptureAccess.preflight [ [Ref](https://github.com/servo/core-foundation-rs/blob/e5162be4fa50c9f7f2f41f1d310eb49268c4e88e/core-graphics/src/access.rs#L17) ] returns incorrectly for Intel based 64 bit Macs.

On further inspection, the reason for this is that the external method `CGPreflightScreenCaptureAccess` returns a value which is not 1 ( even if the permissions have been given ). 
From my research, I've found that it returns a number whose lowest 8 bits are 0. So something like the following should work

image

That's what I'm doing in my fork [here](https://github.com/Pranav2612000/core-foundation-rs)

This is working good enough for me. What do you think about this approach? Does it make sense? I'm open to submitting a PR, or if you have a better solution in mind, I'll love to implement it.

Pranav2612000:
Fixes #677

### DIFF
--- a/core-graphics/src/access.rs
+++ b/core-graphics/src/access.rs
@@ -14,7 +14,7 @@ impl ScreenCaptureAccess {
     /// Return `true` if has access
     #[inline]
     pub fn preflight(&self) -> bool {
-        unsafe { CGPreflightScreenCaptureAccess() == 1 }
+        unsafe { (CGPreflightScreenCaptureAccess() & 1) == 1 }
     }
 }
 
--- a/core-graphics/Cargo.toml
+++ b/core-graphics/Cargo.toml
@@ -1,10 +1,10 @@
 [package]
-name = "core-graphics"
-description = "Bindings to Core Graphics for macOS"
+name = "core-graphics-helmer-fork"
+description = "Fork of core-graphics maintained by helmer team"
 homepage = "https://github.com/servo/core-foundation-rs"
 repository = "https://github.com/servo/core-foundation-rs"
 version = "0.24.0"
-authors = ["The Servo Project Developers"]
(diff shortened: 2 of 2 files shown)


Pranav2612000:
@jrmuizel Can you please take a look?

core-graphics/src/access.rs
```diff
@@ -14,7 +14,7 @@ impl ScreenCaptureAccess {
     /// Return `true` if has access
     #[inline]
     pub fn preflight(&self) -> bool {
-        unsafe { CGPreflightScreenCaptureAccess() == 1 }
+        unsafe { (CGPreflightScreenCaptureAccess() & 1) == 1 }
```
jrmuizel, line 17: Won't this return false if CGPreflightScreenCaptureAccess() returns 2?
Should request() also be changed.

Can you also add some more documentation about what semantics you expect CGPreflightScreenCaptureAccess() to have. i.e. what are the non 1 values you see it returning.

Pranav2612000: > i.e. what are the non 1 values you see it returning.

I don't remember the exact value, but it was a 16bit number where the last 8 bits were `00000001`

> Won't this return false if CGPreflightScreenCaptureAccess() returns 2?

Yes. It would. But I think 2 won't be returned.

>Should request() also be changed.

Yup. I'll do the change if this change looks good. `preflight` was on a higher priority so I did this change first.

waywardmonkeys:
This was fixed in #698. Thanks for your report!

PR #642, opened 2023-10-10
Update to 2018 edition

waywardmonkeys:
This doesn't get away from `extern crate` yet.

Is this set of changes welcome?

PR #683, opened 2024-05-30
Bump version of core-foundation-sys to 0.8.7

wusyong:
I opened a [topic](https://servo.zulipchat.com/#narrow/stream/263398-general/topic/Crate.20maintenance.3A.20core-foundation-rs/near/441237419) in Zulip regarding some crate maintenance under Servo org. @jdm gave some advice on how to update crate version. I guess I can start with sys crate and open other PRs one by one. Does this look good to other maintainers?


jdm:
Which commit in https://github.com/servo/core-foundation-rs/commits/main/core-foundation-sys since the 0.8.6 version update contains breaking changes?

wusyong:
I see #642 as a breaking change since I remember cases of build failure when upgrading from 2015 to 2018.
But in general, I think it will be up to the author to decide. If everyone prefers, I can change it to patch bump instead.

jdm:
Edition changes should not affect the public API. I believe a patch version update would be correct here.

mrobinson (approved):

PR #691, opened 2024-07-20
core-foundation: Use dep syntax for `with-uuid` feature.

waywardmonkeys:
This removes an implicit feature due to the optional dependency `uuid`.

While this is user visible, any use of `uuid` as a feature previously would not have enabled the support code and so it wouldn't have had any benefit.

Bump the MSRV in CI to 1.64 (while only 1.60 is required here, other features from 1.64 will be used).

### DIFF
--- a/core-foundation/Cargo.toml
+++ b/core-foundation/Cargo.toml
@@ -25,7 +25,7 @@ default = ["link"]
 
 mac_os_10_7_support = ["core-foundation-sys/mac_os_10_7_support"] # backwards compatibility
 mac_os_10_8_features = ["core-foundation-sys/mac_os_10_8_features"] # enables new features
-with-uuid = ["uuid"]
+with-uuid = ["dep:uuid"]
 # Disable to manually link. Enabled by default.
 link = ["core-foundation-sys/link"]
 


waywardmonkeys:
Dep syntax needs MSRV 1.60 or later.

mrobinson (approved):

mrobinson:
You'll also need to update the MSRV in this change in the `Cargo.toml` (if it isn't already defined) and fix the CI job.

waywardmonkeys:
@mrobinson Any preference what it gets updated to?

mrobinson:
> @mrobinson Any preference what it gets updated to?

I suppose we should shoot for the oldest version possible.

waywardmonkeys:
@mrobinson  For now, I only bumped it in the CI job. I'd rather do a full separate PR that fixes up the package info to include the `rust-version` separately if that's okay.

waywardmonkeys:
Also, I've bumped to 1.64 rather than 1.60 as that'll get us the features for #692 and also workspace dependencies and package settings.

waywardmonkeys:
The most common dependency of this crate is `security-framework` and I asked over there about their MSRV as they claim 1.60, but actually test for 1.65:

https://github.com/kornelski/rust-security-framework/issues/208

mrobinson:
Okay. 1.65 seems reasonable.

waywardmonkeys:
@mrobinson Would've been nice to have gotten a heads up that a new version was going to be published, including a breaking semver and we could've gotten this landed ... Not sure why I bother with any of these PRs.

waywardmonkeys:
@jdm @mrobinson This too is good to go for review.

jdm (approved):

PR #692, opened 2024-07-23
Use `core::ffi` over `std::os::raw` and `libc::c_*`

waywardmonkeys:
`core::ffi` is the preferred way to access these type definitions. Begin removing uses of the versions from `libc`.

`libc::c_void` is not exactly the same as `core::ffi::c_void` but some code has tried using them interchangeably (like in io-surface).

`libc::size_t` is `usize` on all current platforms.


waywardmonkeys:
This will need Rust 1.64 ...

waywardmonkeys:
@jdm @mrobinson This is good for review now.

jdm (approved):

PR #694, opened 2024-08-11
Bump MSRV to 1.65

waywardmonkeys:
This will allow us to use `dep:` syntax and also move to using types from `core::ffi` as well as some other more modern constructs.


waywardmonkeys:
This is as discussed in #691 and also needed for #692.

I've also discussed this with the maintainer of the `objc2` family of crates.

jdm (approved):

PR #700, opened 2024-08-12
Remove `core_graphics::private::CGSRegion`

### LINKED ISSUE #656: Usage of `CGSRegionRelease` in `core-graphics` prevents publishing to the Mac AppStore
My app got rejected from the Mac AppStore with the following message:
```
Guideline 2.5.1 - Performance - Software Requirements

Your app uses or references the following non-public or deprecated APIs:

Contents/MacOS/parquetviewer
Symbols:
• _CGSRegionRelease

The use of non-public or deprecated APIs is not permitted on the App Store, as they can lead to a poor user experience should these APIs change and are otherwise not supported on Apple platforms.
```

I believe the offending code is here: https://github.com/servo/core-foundation-rs/blob/0933af08be6cf1ba5ad3b0abfc05bbead489116b/core-graphics/src/private.rs#L24

waywardmonkeys:
This is a private API on macOS. This code was added for use in `servo-glutin` and was not used outside of that as far as I know (by doing various code searches).

Fixes #656.


waywardmonkeys:
If there is someone using this, they can take the code that used to be here and use it within their own codebase rather than us maintaining it.

jdm (approved):

mrobinson:
Perhaps we can do a quick turnaround here and release the 0.11 series in order to not block the pending PRs that should be merged?

waywardmonkeys:
I want to go through more of the backlog of things to make sure that we're not missing things. There's at least a couple that should land! Taking a couple of days for that seems fine...

Propagating all of this stuff into the broader ecosystem is going to be some serious work.

PR #701, opened 2024-08-12
Use workspace inheritance.

waywardmonkeys:
Consolidate package configuration where possible. This also removes the `homepage` key as it only points to the repository so isn't providing anything additional.

Use workspace dependencies for the crates provided by the workspace and used within the workspace. This centralizes the version number maintenance a bit.


jdm (approved):
Nice.

PR #708, opened 2024-08-30
core-graphics: move private APIs behind a `private-apis` feature flag, closes #707

### LINKED ISSUE #707: core-graphics: non-public API usage _CGContextSetFontSmoothingStyle
I'm trying to submit an iOS app but Apple is rejecting it:
```
ITMS-90338: Non-public API usage - The app references non-public symbols in <appName>: _CGContextSetFontSmoothingStyle. If method names in your source code match the private Apple APIs listed above, altering your method names will help prevent this app from being flagged in future submissions. In addition, note that one or more of the above APIs may be located in a static library that was included with your app. If so, they must be removed. For further information, visit the Technical Support Information at http://developer.apple.com/support/technical/
```

Even though my dependencies do not use the `set_font_smoothing_style` API, seems like Rust cannot strip it.

Similar to #656

I thought about deleting the API similar to  #700 but seems like at least alacritty uses it. Maybe it makes sense to use a feature flag instead?

lucasfernog-crabnebula:
The `CGContextSetFontSmoothingStyle` API is private, and Apple rejects it for the App Store.

See #707 for more information.

### DIFF
--- a/core-graphics/Cargo.toml
+++ b/core-graphics/Cargo.toml
@@ -28,5 +28,6 @@ libc = "0.2"
 default = ["link"]
 elcapitan = []
 highsierra = []
+private-apis = []
 # Disable to manually link. Enabled by default.
 link = ["core-foundation/link", "core-graphics-types/link"]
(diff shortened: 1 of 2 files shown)


lucasfernog-crabnebula:
closing in favor of #709

mrobinson:
Instead of adding a feature, I think I would prefer to simply remove this API completely as discussed on #707.

PR #712, opened 2024-10-12
core-graphics: Add CGKeyCodes for numpad keys

pentamassiv:
The numpad keys also always have the same values

### DIFF
--- a/core-graphics/src/event.rs
+++ b/core-graphics/src/event.rs
@@ -57,7 +57,7 @@ bitflags! {
 /// layouts may have the 'A' key label on a different physical key;
 /// in this case, pressing 'A' will generate a different virtual
 /// keycode. Constants with the 'JIS_' or 'ISO_' prefix behave
-/// analogous. Keys without a prefix are independent of the
+/// analogously. Keys without a prefix are independent of the
 /// keyboard layout.
 ///
 /// [Ref](https://github.com/phracker/MacOSX-SDKs/blob/master/MacOSX10.13.sdk/System/Library/Frameworks/Carbon.framework/Versions/A/Frameworks/HIToolbox.framework/Versions/A/Headers/Events.h#L197-L327)
--- a/core-graphics/src/event.rs
+++ b/core-graphics/src/event.rs
@@ -43,12 +43,92 @@ bitflags! {
     }
 }
 
-/// Key codes for keys that are independent of keyboard layout.
+/// Constants for the virtual key codes
 ///
-/// [Ref](https://github.com/phracker/MacOSX-SDKs/blob/master/MacOSX10.13.sdk/System/Library/Frameworks/Carbon.framework/Versions/A/Frameworks/HIToolbox.framework/Versions/A/Headers/Events.h)
+/// These constants are the virtual keycodes defined originally in
+/// Inside Mac Volume V, pg. V-191. They identify physical keys on a
+/// keyboard. The struct contains the values of the ANSIKeyCode,
+/// KeyCode, ISOKeyCode and JISKeyCode of the original Carbon headers.
+///
+/// Those constants with "ANSI" in the name are labeled
+/// according to the key position on an ANSI-standard US keyboard.
+/// For example, `ANSI_A` indicates the virtual keycode for the key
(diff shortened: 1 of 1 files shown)


core-graphics/src/event.rs
```diff
@@ -97,6 +98,25 @@ impl KeyCode {
     pub const RIGHT_ARROW: CGKeyCode = 0x7C;
     pub const DOWN_ARROW: CGKeyCode = 0x7D;
     pub const UP_ARROW: CGKeyCode = 0x7E;
+    pub const NUMPAD_0: CGKeyCode = 0x52;
```
wusyong, line 101: Could we use the convention in original [Ref](https://github.com/phracker/MacOSX-SDKs/blob/master/MacOSX10.13.sdk/System/Library/Frameworks/Carbon.framework/Versions/A/Frameworks/HIToolbox.framework/Versions/A/Headers/Events.h)?

mrobinson: I agree, we should use the same name as the original reference. In addition, would it be possible to add *all* the missing key constants here and put them in the order of the original header file with comments indicating what section they came from?

mrobinson (changes requested):

core-graphics/src/event.rs
pentamassiv: I added more constants, but I put them into new structs. Otherwise the comments would have been wrong. I also used the same names as in the original header file.

core-graphics/src/event.rs
```diff
@@ -43,9 +43,90 @@ bitflags! {
...
+/// For example, `ANSI_A` indicates the virtual keycode for the key
+/// with the letter 'A' in the US keyboard layout. Other keyboard
+/// layouts may have the 'A' key label on a different physical key;
+/// in this case, pressing 'A' will generate a different virtual
+/// keycode.
+///
+/// [Ref](https://github.com/phracker/MacOSX-SDKs/blob/master/MacOSX10.13.sdk/System/Library/Frameworks/Carbon.framework/Versions/A/Frameworks/HIToolbox.framework/Versions/A/Headers/Events.h#L197-L261)
+#[repr(C)]
+pub struct ANSIKeyCode;
+impl KeyCode {
```
mrobinson, line 59: Did you meant to write `impl AnsiKeyCode` here? In any case, I think you can just remove the new structures and keep only `KeyCode`, even though they are in separate enums in the Carbon headers, they are part of the same series of values. Separating seems unnecessary here.

mrobinson (changes requested):
Sorry for the long delay. Thanks for included all of the keycodes! I think that only one small set of changes is necessary here.

core-graphics/src/event.rs
pentamassiv: What do you want me to do with the doc comments? I think they add valuable information, but I don't know how to do that if they are not independent structs.  Should they get deleted or merged?

mrobinson: The documentation could be combined into one block for `KeyCode` explaining where all these values come from in CoreGraphics.

pentamassiv: Okay, I combined the comments. Is this how you imagined it?

core-graphics/src/event.rs
```diff
@@ -43,12 +43,92 @@ bitflags! {
...
+/// KeyCode, ISOKeyCode and JISKeyCode of the original Carbon headers.
+///
+/// Those constants with "ANSI" in the name are labeled
+/// according to the key position on an ANSI-standard US keyboard.
+/// For example, `ANSI_A` indicates the virtual keycode for the key
+/// with the letter 'A' in the US keyboard layout. Other keyboard
+/// layouts may have the 'A' key label on a different physical key;
+/// in this case, pressing 'A' will generate a different virtual
+/// keycode. Constants with the 'JIS_' or 'ISO_' prefix behave
+/// analogous. Keys without a prefix are independent of the
```
mrobinson, line 60:
```suggestion
/// analogously. Keys without a prefix are independent of the
```

mrobinson (approved):
Yes. Thank you!

PR #713, opened 2024-10-14
core-graphics: add event set_location

iparaskev:


core-graphics/src/event.rs
```diff
@@ -868,4 +874,5 @@ extern "C" {
 
     fn CGEventTapEnable(tap: CFMachPortRef, enable: bool);
 
+    fn CGEventSetLocation(event: crate::sys::CGEventRef, location: CGPoint);
```
mrobinson, line 877: All other `CG...` functions here have rust documentation, so it makes sense to add it here as well, I think.

iparaskev: Good point, I added to `CGEventTapEnable` as well.

mrobinson:
Thanks for the contribution!

PR #714, opened 2024-11-12
fix: update NSWindowTabbingMode enum values

gaauwe:
The enum values for `NSWindowTabbingModeDisallowed` and `NSWindowTabbingModePreferred` are incorrect, according to https://developer.apple.com/documentation/appkit/nswindow/tabbingmode


madsmtm (approved):

jdm (approved):

gaauwe:
Great to see that it got merged so quickly! 🙏 

Maybe I overlooked some documentation, but what is the release process? I have an open PR for another project that uses the `NSWindowTabbingMode`, so it would be great if we can use the correct values 😄

PR #716, opened 2025-01-04
core-graphics: add binding for CGColorCreateSRGB

Jay-Madden:
Very few of the C methods for colors are included here, this one is a useful one to have implemented.

https://developer.apple.com/documentation/coregraphics/3042355-cgcolorcreatesrgb?language=objc

### DIFF
--- a/core-graphics/src/color.rs
+++ b/core-graphics/src/color.rs
@@ -27,6 +27,13 @@ impl CGColor {
             CGColor::wrap_under_create_rule(ptr)
         }
     }
+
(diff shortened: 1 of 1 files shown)


jrmuizel:
It seems like this first showed up in 10.15 perhaps it's inclusion should be made conditional behind a `catalina` feature?

mrobinson:
> I can add that if needed

Yes, please create a PR to do that. Thank you and sorry for not catching that earlier!

Jay-Madden:
@mrobinson @jrmuizel  Finally got around to this, opened #719

PR #719, opened 2025-01-18
core-graphics: add catalina feature flag for srgb cgcolor binding

Jay-Madden:


mrobinson (approved):
Thanks so much for this follow up!

PR #723, opened 2025-03-16
Fix CI

tmandry:
For the check-cfg workaround see https://github.com/SSheldon/rust-objc/issues/125. This is the smallest change that will fix CI. Alternatives include switching to the objc-rs fork and switching to objc2, both of which would be more disruptive and require breaking changes.

Using `-Dwarnings` in CI is probably not a good idea without pinning a toolchain version, because warnings get added in newer Rust versions all the time, but this PR doesn't change that yet.

Semver checks must have been failing when they were enabled. I allowed failures for now until all the versions are bumped (this will be transitioning from bump-at-release to bump-at-first-breaking-change).

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -27,6 +27,8 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     - name: Check semver
+      # Allow failure until we update all the package versions.
+      continue-on-error: true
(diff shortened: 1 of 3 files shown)


tmandry:
Not sure what to do about the semver checks, those must have been failing when they were turned on. The check should probably be marked optional.

jdm (approved):
Thank you!

PR #724, opened 2025-03-16
[breaking][core-graphics] Allow dropping event from CGEventTap callback

tmandry:
This allows CGEventTap callbacks to capture (drop) an event entirely. This is useful for implementing system-wide hotkeys, among other things.

This PR is essentially a rebase of #492. While releasing the breaking change in #710 we might as well include this, since users of CGEventTap will need to update their code anyway.

Closes #492, #664, #665.

PR #725, opened 2025-03-16
[breaking][core-graphics] Enforce Sendness of CGEventTap callback (soundness fix)

tmandry:
There was an aspect of unsoundness missed in #710. The callback can capture state from the current thread that is not `Send`, but the tap can be installed on another thread's event loop.

This change fixes the unsoundness while providing a way to safely and conveniently use the API that fits most use cases. `CGEventTap::with` (unreleased API) was renamed to `with_enabled` because it actually installs and enables the tap now.

Also make CGEventTap `#[must_use]` so users of `new` don't accidentally drop the tap.

This change is currently based on #724 but can be rebased if you want to merge it independently.

### DIFF
--- a/core-graphics/src/event.rs
+++ b/core-graphics/src/event.rs
@@ -567,6 +567,7 @@ unsafe extern "C" fn cg_event_tap_callback_internal(
 ///     },
 /// ).expect("Failed to install event tap");
 /// ```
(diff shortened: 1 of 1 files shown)


tmandry:
cc @wusyong

jdm (approved):

Issue #729, opened 2025-04-20
Future of this project and discussion about `objc2`


madsmtm:
Hey @waywardmonkeys, @jrmuizel, @mrobinson and @wusyong (and CC @tmandry and @simlay).

Over the past few years, I've been working on replacements for the crates in this repo in [the `objc2` project](https://github.com/madsmtm/objc2), see https://github.com/servo/core-foundation-rs/pull/513 and https://github.com/servo/core-foundation-rs/pull/628 for some previous discussion. To rehash the benefits of the `objc2-*` crates:
- Code generation from headers, which means that basically every upstream API is available (only a few explicitly skipped APIs aren't), and this will remain true even as Apple releases new SDK versions over time (since we only need to rerun the generator).
- Automatic memory management, all objects are wrapped by the code generator in either `objc2::rc::Retained` or `objc2_core_foundation::CFRetained`, which will release the object on `Drop`. The generator knows about the "create rule" and about opt-outs like `CF_RETURNS_NOT_RETAINED` etc. (so it will do this correctly with a much higher probability than if a human did it).
- Type-safety, especially compared to `cocoa` and `cocoa-foundation` where everything is just an extension trait on the `id` pointer.

I'm motivated in bringing these benefits to the entirety of the Apple+Rust ecosystem, as I think the soundness, correctness and ergonomics improvements here are quite significant. To that end, I've been slowly migrating major Rust projects, see https://github.com/madsmtm/objc2/issues/174.

A recent milestone is that I've released `objc2-core-foundation v0.3.1`, which I believe now covers everything that `core-foundation` does! I've also assessed every open issue and pull-request in this repo, and made sure that they're either resolved or tracked in the `objc2` repo, see https://github.com/madsmtm/objc2/issues/719. So at this point, I'd like to start discussing the future of this project, and how we can migrate Servo and other users to the `objc2-*` crates (assuming you agree that's a goal, if not then I'd also like to discuss that!)

My ideal scenario would be that the crates in this repo were marked `#![deprecated]` with re-exports of the `objc2-*` variants, both to give users an easier migration path, and because in a way it'd "bless" the `objc2` crates, but I can see other solutions, including any of the following variations:
- Only do this for `cocoa` and `cocoa-foundation` (these are where `objc2-*` would be the most valuable).
- Soft-deprecate the crates in the README only, but without actually marking it `#![deprecated]`.
[...]

PR #731, opened 2025-05-12
Deprecate `cocoa`, `cocoa-foundation` and `io-surface` and link to `objc2` crates instead

madsmtm:
Resolves part of https://github.com/servo/core-foundation-rs/issues/729, see that for motivation.

I believe it should be safe to move forwards with deprecating `cocoa` and `cocoa-foundation`, their major reverse dependencies have [already migrated](https://github.com/madsmtm/objc2/issues/174), so the ecosystem disruption nowadays will be as minimal as possible. For `io-surface`, the only real dependency seems to be `surfman`, and I have filed https://github.com/servo/surfman/pull/334 for migrating that.

I'll wait with the other crates for now.


cocoa/src/lib.rs
```diff
@@ -1,3 +1,4 @@
+//! This crate has been deprecated in favour of the `objc2` crates.
```
mrobinson, line 1: Please move this below the copyright notice.

io-surface/src/lib.rs
```diff
@@ -1,3 +1,4 @@
+//! This crate has been deprecated in favour of the `objc2-io-surface` crate.
```
mrobinson, line 1: Please move this below the copyright notice.

mrobinson (approved):
Looks good, with a couple small changes. Sorry for the delay here.

madsmtm:
Do you want me to do a version bump PR for it?

madsmtm:
I opened https://github.com/servo/core-foundation-rs/pull/732 for that.

PR #735, opened 2025-07-12
Fix unwrap panic in `extract_number_for_key` in `impl TraitAccessorPrivate for CTFontTraits` on macOS Ventura.

vorporeal:
This splits out the final chunk of https://github.com/servo/core-foundation-rs/pull/517 to be reviewed and merged.

As discussed on that PR, in an ideal world, we'd have a test we could write to prove that this is required.

In practice, Warp has been using the logic in that PR for the past 3 years without issues.  We did attempt, 6 months ago, to switch over to upstream (without the patch here) and had to revert that change due to this still being an issue for users running Ventura.

Given the above, we have confidence that:
1. This patch is still required for proper behavior on macOS Ventura, and
2. It is sufficiently well-tested, in production, that it is safe to merge.

### DIFF
--- a/core-text/src/font_descriptor.rs
+++ b/core-text/src/font_descriptor.rs
@@ -19,8 +19,8 @@ use core_foundation::url::{CFURLRef, CFURL};
 use core_foundation::{declare_TCFType, impl_CFTypeDescription, impl_TCFType};
 use core_graphics::base::CGFloat;
 
-use std::path::PathBuf;
 use core_foundation::boolean::CFBoolean;
+use std::path::PathBuf;
 
 /*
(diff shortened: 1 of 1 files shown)


jdm (approved):

jdm:
Yeah, I think those are clippy failures from newer rust versions that we need to address before we can merge anything.

jdm:
Please go ahead and rebase to have the CI issues go away.

vorporeal:
Done.

vorporeal:
No rush at all (we are using a `[patch]` directive), but would be awesome if this could go out as a `core-foundation 0.10.2` release at some point. :)

PR #736, opened 2025-07-12
Fix clippy lints.

jdm:


Taym95 (approved):
A few more Clippy warnings need to be fixed.

PR #741, opened 2025-09-19
Fix use-after-free issues in Core Foundation object management

quantpoet:
Fixed some potential use-after-free bugs identified by static analysis in the core-text crate. The issues were caused by improper lifetime management of Core Foundation objects where raw pointers were being used after the underlying objects were dropped. Specifically:

In CTFontDescriptor::traits(): Eliminated intermediate CFType wrapper that was being dropped before the dictionary was returned

In create_font_descriptor() functions: Ensured CFData objects remain alive during C function calls by restructuring the unsafe block

These changes maintain the same API while ensuring memory safety through proper Core Foundation reference counting.


quantpoet:
@jdm @Taym95 
Hi, Could you please review this PR at your convenience? Thank you very much.

PR #744, opened 2025-11-11
CI: macos-13 -> macos-15-intel

sagudev:

PR #747, opened 2026-01-12
Upgrade MSRV to 1.68 and remove `macos-13` builder

mrobinson:
This should fix the build on CI. This also incorporate #744
as that is required to get the CI working again.

Signed-off-by: Martin Robinson <mrobinson@igalia.com>

### DIFF
--- a/core-foundation/src/filedescriptor.rs
+++ b/core-foundation/src/filedescriptor.rs
@@ -134,6 +134,7 @@ mod test {
     }
 
     #[test]
+    #[allow(unused_assignments)]
     fn test_callback() {
         let mut info = TestInfo { value: 0 };
         let context = CFFileDescriptorContext {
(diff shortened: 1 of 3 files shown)


mrobinson:
@sagudev I've rolled your change from #744 into this one in order to fix the CI. Do you mind taking a look?

core-foundation/src/filedescriptor.rs
```diff
@@ -169,7 +169,6 @@ mod test {
         CFRunLoop::run_current();
         assert_eq!(info.value, kCFFileDescriptorWriteCallBack);
 
-        info.value = 0;
```
sagudev, line 172: Not sure if this assignment is really unused, given that we have ptr to info ` info: &mut info as *mut _ as *mut c_void,`, although given aliasing rules I get why compiler says so, but that's not what it was probably intended. The correct thing would be to use UnsafeCell or smth like that here.

So I think we should keep it to preserve original intention.

sagudev (approved):
otherwise LGTM

